PR #33, opened 2019-11-15
configurable HTTP parameters

bradymadams:
@thopiekar I made some changes to allow us to configure the HTTP parameters because we'll need to run against a local running server when Rick and Doug are doing demos at Formnext next week, because they won't have internet access.

Would you mind reviewing what I've done to make sure I haven't introduced a bug. I think it's solid and I tested it on my end, but I'm new to working on Cura plug-ins :slightly_smiling_face:


bradymadams:
Oh and I also adjusted the `_wait_time`. I think 50 ms will cause a lot of HTTP requests, and we pay per request on AWS, so I just want to reduce that a bit.

SmartSlicePlugin/SmartSliceExtension/SmartSliceCloudConnector.py
```diff
@@ -118,13 +118,37 @@ def __init__(self, connector) -> None:
...
                                        }
 
+        # default connection settings
+        protocol = 'https'
+        hostname = 'api-19.fea.cloud'
+        port = 443
+
+        # see if we have a config file where we can grab the HTTP client settings from
+        this_dir = os.path.split(__file__)[0]
+        config_location = os.path.join(this_dir, "data", "config.json")
```
thopiekar: ~What is inside the config.json?~ Can't we use Cura's preference storage for that? :thinking:

bradymadams: Probably. I'm unfamiliar with it. Any pointers on where I can look to learn more about it?

thopiekar: Search the same source for "smartslice/token". I was planning to save the token there. In the init of the extension the last argument is the default value.

PR #57, opened 2019-12-09
Boundary Conditions are Validated

CodyFagley:
Validate button stays inactive until a load/anchor are *both* applied.   

Messages update accordingly, but will be ultimately changed to match Teton Mockups.


SmartSlicePlugin/SmartSliceSelectTool/SmartSliceSelectTool.py
```diff
@@ -137,18 +137,37 @@ def _onSelectedFaceChanged(self):
             Logger.log("d", "loaded_faces: {}".format(loaded_faces))
             Logger.log("d", "anchored_faces: {}".format(anchored_faces))
             
-            cloud_connector = PluginRegistry.getInstance().getPluginObject("SmartSliceExtension").cloud
             if self._selection_mode is SelectionMode.AnchorMode:
+                print ("\n\nANCHOR APPLIED RIGHT HERE\n\n")
```
bradymadams: Is this print statement (and the one below) for debugging? Or do we want to keep it?

SmartSlicePlugin/SmartSliceExtension/SmartSliceCloudProxy.py
```diff
@@ -18,13 +18,14 @@
 class SmartSliceCloudStatus:
```
bradymadams: We should inherit from `enum.Enum`

bradymadams:
Accidentally deleted the source branch which closed the PR. Branch has been restored and PR re-opened.

SmartSlicePlugin/SmartSliceSelectTool/SmartSliceSelectTool.py
CodyFagley: This has been removed as part of the nextmerge request, I believe.

SmartSlicePlugin/SmartSliceExtension/SmartSliceCloudProxy.py
CodyFagley: Confirmed; I just read up on this.  I will switch our enumerations to enum.Enum as part of best practice.

Issue #30, opened 2019-11-15
Draw arrow for Load normal to selected surface


dougkenik:

dougkenik:
@aklempau - some issues I've noticed.

IN general, requirements for this ticket are [here](https://tetoncomposites.atlassian.net/wiki/spaces/CURA/pages/99582035/Load+Direction+Arrow). Let's continue to work towards these requirements and acceptance criteria.

1. The arrow is not being centered on the face. It appears you might be computing the center for one triangle, but not the aggregation of triangles which define the face. As an example, the center for the below face is at the center of the square.

![image](https://user-images.githubusercontent.com/24813209/73409804-896b8a80-42bd-11ea-9f86-580952fee625.png)

2. "Flip direction" is not working. If "flip direction" is not selected in the Load Dialogue, it should default to the arrow pointing towards the surface. If the user selects the "flip direction" radio button, it should change to pointing out

3. Color and size of the arrow need to change a bit. Something like this:
![image](https://user-images.githubusercontent.com/24813209/73410017-36de9e00-42be-11ea-8dcb-ef16e924a8c8.png)

I don't have a good idea of what "blue" we want yet (and I don't expect the shading), but if we should elongate the tail of the arrow and shorten the arrow head to match the picture.

aklempau:
For "Flip Direction" radio button, should the arrow simply be drawn reversed by 180 degrees, or is it more involved than that?  Should the triangles themselves be drawn in a counterclockwise manner so that the normal direction of the triangle is changed?

dougkenik:
Simply flipping by 180 will suffice. I don't think you need to redraw the triangles

aklempau:
Pushed new changes to r20.0-load-normal-arrow branch.  The arrow is now centered on face.  If center is not touching portion of face, then arrow is centered over random triangle.  The flip direction checkbox is now working.

dougkenik:
This is in PR #75

dougkenik:
@aklempau - we also need to check that the arrow and force vector which is being sent with the smart slice job are pointed in the same direction, if you haven't already

PR #75, opened 2020-01-29
Add normal arrow to load face.

aklempau:
The normal arrow looks like a spear.  Centering on the face can be difficult given odd shapes.  It is colored green.  I am open to suggestions.


SmartSlicePlugin/SmartSliceSelectTool/SmartSliceSelectHandle.py
```diff
@@ -178,12 +179,57 @@ def drawFaceSelection(self, mode, selmesh : SelectableMesh, draw_arrow = False):
...
+        p_base1 = Vector(center.x, center.y + size, center.z)
+        p_base2 = Vector(center.x, center.y - size, center.z)
+        p_base3 = Vector(center.x + size, center.y, center.z)
+        p_base4 = Vector(center.x - size, center.y, center.z)
+        p_base5 = Vector(center.x, center.y, center.z + size)
+        p_base6 = Vector(center.x, center.y, center.z - size)
+        
+        mb.addFace(p_base2, p_tip, p_base1, color=self.ZAxisSelectionColor)
+        mb.addFace(p_base3, p_tip, p_base4, color=self.ZAxisSelectionColor)
+        mb.addFace(p_base6, p_tip, p_base5, color=self.ZAxisSelectionColor)
```
bradymadams: Add blank line between functions.

aklempau: Blank line has been added.  Added new methods, and put blank lines around those as well.

dougkenik:
Ok, this looks really good! 

I think we need a little longer tail on the arrow to make it a little more easy to see the direction

thopiekar:
Don't know how much it has been discussed internally, but I would recommend giving the arrow some volume. Right now the arrow looks like this:
image

However, I would recommend a mesh looking like this:
Suggestion

Could be also that this is my personal taste of course. :)

Apart from that the code looks clean, newly added functions include descriptions also, variablenames are good to follow and everything is easy to follow within 3 commit and a small set of changes.

Good work!

dougkenik:
@thopiekar I think you're right. Just some things we need to work through for look and feel.In posted some ideas in the original issue #30

aklempau:
Found small issue with Flip Direction.  When arrow exists, then body is deselected and then re-selected (activated), the Flip Direction check box stops working.  The surface needs to be clicked again for the Flip Direction is start working.  I'm looking into this bug down now.

dougkenik:
@aklempau - we also need to check that the arrow and force vector which is being sent with the smart slice job are pointed in the same direction, if you haven't already

dougkenik:
Ok, I'm happy with the look and feel of the arrow. Nice job!

SmartSlicePlugin/select_tool/FaceSelection.py
```diff
@@ -203,6 +203,44 @@ def generateNormalVector(self):
...
+        Inputs:
+            point (SelectablePoint) is single point.
+        Output:
+            Boolean (bool) True means point lies within face. False means point is outside of face.
+        Notes:
+            The tolerance expansion near end makes this method not 100% accurate.
+        """
+        for i, p1 in enumerate(self._points[:-2]):
+            p2 = self._points[i + 1]
+            p3 = self._points[i + 2]
```
bradymadams: From my observation it looks like `len(self._points)` is always 3, so is there a purpose for the enumerator that is not obvious? Would this be equivalent?

```python
p1 = self._points[0]
p2 = self._points[1]
p3 = self._points[2]
```

SmartSlicePlugin/select_tool/SmartSliceSelectHandle.py
```diff
@@ -178,12 +181,141 @@ def drawFaceSelection(self, mode, selmesh : SelectableMesh, draw_arrow = False):
...
+        mb.addFace(p_base5, p_head, p_base1, color=self.YAxisSelectionColor)
+        mb.addFace(p_base6, p_head, p_base1, color=self.YAxisSelectionColor)
+        mb.addFace(p_base6, p_head, p_base2, color=self.YAxisSelectionColor)
+        mb.addFace(p_base2, p_head, p_base5, color=self.YAxisSelectionColor)
+        mb.addFace(p_base3, p_head, p_base5, color=self.YAxisSelectionColor)
+        mb.addFace(p_base5, p_head, p_base4, color=self.YAxisSelectionColor)
+        mb.addFace(p_base4, p_head, p_base6, color=self.YAxisSelectionColor)
+        mb.addFace(p_base6, p_head, p_base3, color=self.YAxisSelectionColor)
+        
+        size = 0.8
```
bradymadams: Do you mind moving the various sizes to be class variables, and renaming them to indicate what part of the arrow they are sizing? This will make it easier to adjust the arrow dimensions in the future, if necessary.

SmartSlicePlugin/select_tool/FaceSelection.py
aklempau: self._points can be a long list of points.  I believe it contains all points in the face.
This method goes through each triangle in face to determine whether input point is inside each triangle.  It checks all triangles using the for loop.
Is len(self._points) always a multiple of 3?  If so, then perhaps the code checks too many triangles.

SmartSlicePlugin/select_tool/SmartSliceSelectHandle.py
aklempau: Good suggestion.  Will do.

SmartSlicePlugin/select_tool/FaceSelection.py
bradymadams: Okay, I must have interpreted something incorrectly. We can leave it as is to be safe.

PR #89, opened 2020-02-23
Multiple printable body check

### LINKED ISSUE #42: Disable shadow on bottom of part
I don't think we need the shadow on the bottom of the part.

### LINKED ISSUE #72: Multiple Body Error
If a user brings in more than 1 Normal mesh on a build plate and navigates to Smart Slice, we need to issue an error and not display any bodies in the screen.

We will except 1 Normal mesh and multiple Modifier meshes. 

Mockups with workflow description below (workflow is in speaker notes)

https://tetoncompositescom.sharepoint.com/:p:/s/TetonSimulation/EcElKda05p5EiPhhMqXmPDgB43oWVscDqKtwBqlkLrQaJA?e=6h4hku

bradymadams:
Resolves #72 
Resolves #42 

This implements a check for how many "printable meshes" are in the scene when Smart Slice is selected. If it is anything other than 1 a message is shown and the user is sent back to the Prepare Stage.

Additionally, I created a custom View for SmartSlice to have better control over the rendered SceneNodes. It looks and behaves a little nicer now.


bradymadams:
@dougkenik please try this build when you get a chance and let me know if anything should be changed.

https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/SmartSlicePlugin-r20.0-multiple-body-error.49.curapackage

dougkenik:
If I open a model after entering Smart Slice, I do get a warning in the bottom right, but I'm thinking we should be issuing the same error as in prepare. Thoughts?

image

bradymadams:
That is not the correct error or behavior I implemented. It shouldn't even be entering the Smart Slice stage. Are you sure you installed the correct plug-in? If so, I may need to understand more about what you are doing, because it was very stable in the ways that I tested it - I never saw it misbehave.

dougkenik:
To reproduce, I opened one model, went to Smart Slice, and opened another

bradymadams:
> To reproduce, I opened one model, went to Smart Slice, and opened another

It should have been, but it wasn't obvious to me that one could load models within the Smart Slice stage, so I hadn't tested that. This should work better for that scenario:

https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/SmartSlicePlugin-r20.0-multiple-body-error.55.curapackage

dougkenik:
Looks much better. I think this is good fore the Alpha

Issue #97, opened 2020-03-17
Smart Slice fails after a change in printer


bradymadams:
[Originally reported](https://trello.com/c/A81ug0UQ/23-error-after-switching-printer) by @rickdalgarno

I opened Cura and an S5 with Breakaway material was the current printer/material combo. I then switched to an S3 and Cura changed the material to ABS. I ran a SS validation and it resulted in the error shown in the attached image.

I suspect that the material did not update in SS when the printer was switched from the S5 to the S3. 

Steps to reproduce:
1. Open Cura and select an S5 printer and Breakaway as the material.
2. Switch the printer to an S3.
3. Open the pedal fixture and setup a validation step.
4. Click Validate and note the error.
See attached video for this workflow.

Probably not critical for alpha but something that needs to be addressed.

![image](https://user-images.githubusercontent.com/23110600/76896011-b74d6400-6856-11ea-8d72-9473edbb5118.png)

bradymadams:
Done in #112

PR #112, opened 2020-04-16
Update SmartSliceCloudConnector.py

jake-kirkland-teton:
Simple change. There is a connection added to monitor if the printer changes, and then the SS Connector object refreshes the active machine so it no longer attempts to use the previous printer's materials and nozzle settings.


bradymadams:
Fixes #97

Issue #71, opened 2020-01-22
Extruder Change Warning


dougkenik:
We currently only support the 1st Extruder in a multi-extruder printer.

If the user changes the Extruder for the geometry to anything but Extruder 1, or the user selects the wall or Infill extruder to Extruder 2, we need to submit a warning to the user (see mockups). The extruder can only be Not Overriden, or Extruder 1

If the user selects Ok, we will change the extruder back to Extruder 1.

Mockups below (Speaker notes have workflow)

https://tetoncompositescom.sharepoint.com/:p:/s/TetonSimulation/ET6obTueAGRKsjXRr4ABqvIBkMhaoeKl7VoYjMN1jBelXg?e=PgtxgZ

bradymadams:
Done in #118

Issue #56, opened 2019-12-07
Pre-validate Slicer Profile


bradymadams:
- [x] Design UX
- [x] Implement

See comments in #118 about moving the extruder checks to the profile validation step.

PR #118, opened 2020-05-04
R20.1 extruder change warning

jake-kirkland-teton:
This resolves issue #71 Extruder Change Warning.

The solution prevents the user from changing to another extruder in the SmartSlice tab, and will change it back to extruder 1 if it was changed before moving to the SmartSlice tab.


bradymadams:
It doesn't appear the current logic will catch extruder assignments for the wall and infill extruder that are not extruder 1, as noted in the #71.

@dougkenik there's a [build of this branch here](https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/SmartSlicePlugin-r20.1-extruder-change-warning.1.curapackage)

Can you check it out and give Jake any feedback on the current behavior?

dougkenik:
Yeah, the behavior is fine for changing the selected extruder on the left menu or the top menu, but if it's changed in the settings, it won't catch (picture below). It takes a few seconds for the error to be thrown on my machine. Anyone else seeing that?

image

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -80,12 +88,34 @@ def _exit_stage_if_scene_is_invalid(self):
...
+        machine_extruders = list(filter(
+            lambda extruder: extruder.position == active_extruder_position,
+            CuraApplication.getInstance().getMachineManager().activeMachine.extruderList
+        ))
+
+        if len(machine_extruders) > 0:
+            return machine_extruders[0]
+
+        return None
+
```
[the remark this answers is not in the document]
jake-kirkland-teton: Thanks for the info. I have that extension installed.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -101,6 +131,17 @@ def onStageSelected(self):
         # plugin name.
         controller.setActiveView('SmartSlicePlugin')
 
+        # We currently only support using the first extruder on a given machine, so this
+        # limits the print to only use the first extruder while the user is in the 
+        # SmartSlice tab.
+        self._current_printing_node = getPrintableNodes()[0]
```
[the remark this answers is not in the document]
jake-kirkland-teton: Ahh, understood. I don't know what I was thinking there, for some reason I guess I didn't trace that assignment back to see it was exactly what I was using. Thanks!

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -101,6 +131,17 @@ def onStageSelected(self):
...
+        # We currently only support using the first extruder on a given machine, so this
+        # limits the print to only use the first extruder while the user is in the 
+        # SmartSlice tab.
+        self._current_printing_node = getPrintableNodes()[0]
+        active_extruder = self.getNodeActiveExtruder(self._current_printing_node)
+        emActive = self._extruderManager._active_extruder_index
+        self._current_printing_node.callDecoration("getActiveExtruderChangedSignal").connect(self.showExtruderDialog)
+
+        if int(active_extruder.getMetaDataEntry("position")) == 1 or int(emActive) == 1:
+            self.showExtruderDialog()
```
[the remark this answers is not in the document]
jake-kirkland-teton: I think that will work. I thought about it a bit, and I think that is a good way to resolve that. Since it only happens on tab switch, it should really not be a big waste of clock time to always run the dialog method. And definitely more robust, very good point.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -116,7 +157,41 @@ def onStageSelected(self):
...
+    # first extruder on a machine is currently supported. 
+    def showExtruderDialog(self):
+        active_extruder = self.getNodeActiveExtruder(self._current_printing_node)
+        emActive = self._extruderManager._active_extruder_index
+
+        # If both extruders somehow already are extruder one by the time the dialog
+        # starts to activate, we can skip creating it.
+        if int(active_extruder.getMetaDataEntry("position")) == 0 and int(emActive) == 0:
+            return
+        # Otherwise, if the dialog is already visable, or if the user is no longer on the
```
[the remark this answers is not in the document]
jake-kirkland-teton: Got that changed. Sorry for the sloppiness.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -80,12 +88,34 @@ def _exit_stage_if_scene_is_invalid(self):
             return None
         return printable_nodes[0]
 
+    def getNodeActiveExtruder(self, node : SceneNode) -> ExtruderStack:
```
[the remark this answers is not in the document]
jake-kirkland-teton: I thought about doing this when I copied it and I really should have. I have it moved now!

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -116,7 +131,75 @@ def onStageSelected(self):
 
         self._connector.updateSliceWidget()
 
+    def extruderPropertyChanged(self, key: str, property_name: str):
+        if key == "wall_x_extruder_nr" or key == "wall_0_extruder_nr" or key == "infill_extruder_nr":
+            self.showExtruderDialog()
```
bradymadams, line 136: Do we need to check for `wall_extruder_nr` here?

The more standard way to check if a variable is one of multiple values is to check against a list:

```python
if key in ("wall_x_extruder_nr", "wall_0_extruder_nr", etc....):
```

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -116,7 +131,75 @@ def onStageSelected(self):
...
+        emActive = extruderManager._active_extruder_index
+
+        # If both extruders are extruder one we can skip creating the dialog.
+        if (int(active_extruder.getMetaDataEntry("position")) == 0 and
+            int(emActive) == 0 and
+            int(active_extruder.getProperty("wall_extruder_nr", "value")) <= 0 and
+            int(active_extruder.getProperty("wall_0_extruder_nr", "value")) <= 0 and
+            int(active_extruder.getProperty("wall_x_extruder_nr", "value")) <= 0 and
+            int(active_extruder.getProperty("infill_extruder_nr", "value")) <= 0):
+                return
```
bradymadams, line 153: Extra level of indentation here.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -116,7 +131,75 @@ def onStageSelected(self):
...
+            machine.setProperty("wall_extruder_nr","value",0)
+
+        if int(machine.getProperty("wall_0_extruder_nr","value")) > 0:
+            machine.setProperty("wall_0_extruder_nr","value",0)
+
+        if int(machine.getProperty("wall_x_extruder_nr","value")) > 0:
+            machine.setProperty("wall_x_extruder_nr","value",0)
 
+        if int(machine.getProperty("infill_extruder_nr","value")) > 0:
+            machine.setProperty("infill_extruder_nr","value",0)
```
bradymadams, line 202: Arguments should be separated with a comma and single space.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -116,7 +131,75 @@ def onStageSelected(self):
...
+            lifetime=0
+        )
+
+        # Here we show the popup if it needs displayed and set the extruder 1
+        self.resetToExtruderOne()
+        self._extruderDialog.show()
+
+    # This resets both the side bar buttons for the extruder, as well as the drop down
+    # menus that change the extruder.
+    def resetToExtruderOne(self):
```
bradymadams, line 178: Let's have this take in the `node` as an argument, since it is already being used in the function that calls this function.

How hard do you think it would be to re-work this to take the extruder number to reset to?

```python
def resetExtruder(self, node, extruder_nr : int):
```

My thought process is we will likely support the other extruders in a future release, but may not support mixed extruders. So if this function took the extruder number it will be more flexible for those future use cases.

If you think it will be a pain to update, then we can hold off.

bradymadams:
@jake-kirkland-teton there's also conflicts with the base branch, so you'll need to merge `r20.1` into this branch to resolve those.

bradymadams:
@dougkenik after playing around with this I think maybe we missed something about your mockups. Did you intend for this check to be made when the SmartSlice stage is selected or when Validation/Optimization is clicked?

Currently, the logic is doing the former, which after playing around with it I think results in a poor user experience because we may reset extruders when the user would rather that they not be reset.

I think the latter would be better because it would give the user the option to reset extruders or not. But I also think we need to add a `Cancel` button to the mockups, so they can click `Ok` to proceed, or `Cancel`.

Thoughts?

dougkenik:
I think the latter is preferred, but if we're implementing the pre-slice warning, I think that will resolve all of this. Is it worth the effort of redoing this if it will be redone?

bradymadams:
If you're referring to #56 then you make a good point. Let's make a point to move the functionality in this PR to the slice profile validation step when we get to working on #56. I'll make a note in there so it isn't missed.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -18,11 +18,14 @@
 from UM.PluginRegistry import PluginRegistry
 from UM.Message import Message
 from UM.Scene.Selection import Selection
+from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
 
 from cura.Stages.CuraStage import CuraStage
 from cura.CuraApplication import CuraApplication
+from cura.Settings.SettingOverrideDecorator import SettingOverrideDecorator
```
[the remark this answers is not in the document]
jake-kirkland-teton: Done.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -18,11 +18,14 @@
 from UM.PluginRegistry import PluginRegistry
 from UM.Message import Message
 from UM.Scene.Selection import Selection
+from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
```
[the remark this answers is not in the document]
jake-kirkland-teton: Done.

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: I thought about that check being necessary or not a few times, I will add it, just for completeness. And good point. Changed.

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: I did this for readability. Removed.

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: Done.

jake-kirkland-teton:
OK. I think all the requested changes have been made. Take a look and let me know if you see anything else I should adjust. Thanks for the feedback!

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -127,6 +147,61 @@ def onStageSelected(self):
...
+    def showExtruderDialog(self):
+        current_printing_node = getPrintableNodes()[0]
+        active_extruder = getNodeActiveExtruder(current_printing_node)
+        extruderManager = CuraApplication.getInstance().getExtruderManager()
+        emActive = extruderManager._active_extruder_index
+
+        # If both extruders are extruder one we can skip creating the dialog.
+        if (int(active_extruder.getMetaDataEntry("position")) == 0 and
+        int(emActive) == 0 and
+        all(map(lambda k : (int(active_extruder.getProperty(k,"value")) <= 0), self.EXTRUDER_KEYS))):
```
bradymadams, line 165: space between `k` and `"value"`

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -127,6 +147,61 @@ def onStageSelected(self):
 
         self._connector.updateSliceWidget()
 
+    def extruderPropertyChanged(self, key: str, property_name: str):
+        if key in ("wall_extruder_nr", "wall_x_extruder_nr", "wall_0_extruder_nr", "infill_extruder_nr"):
```
bradymadams, line 151: Use `EXTRUDER_KEYS` here.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -127,6 +147,61 @@ def onStageSelected(self):
...
+            lifetime=0
+        )
+
+        # Here we show the popup if it needs displayed and set the extruder 1
+        self.resetToExtruderOne(current_printing_node, extruderManager, emActive, active_extruder, 0)
+        self._extruderDialog.show()
+
+    # This resets both the side bar buttons for the extruder, as well as the drop down
+    # menus that change the extruder.
+    def resetToExtruderOne(self, node, extruderM, extruderMA, activeE, extruder_nr : int):
```
bradymadams, line 191: Let's rename to `resetExtruder` or something similarly generic since it's no longer restricted to resetting to extruder 0.

Let's use more descriptive names for `extruderM`, `extruderMA`, and `activeE`. It becomes less clear what they're supposed to be once we get into this function.

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: Yeah I saw this one after I committed and felt like another commit for a single space might be kinda silly. Got it.

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: Duh. My bad. Very good point.

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: Very good call. Done.

Fair point. I have changed them to be more explicit.

Issue #85, opened 2020-02-12
Smart Slice fails to load on fresh Cura install


bradymadams:
In a fresh Cura install there is no active machine selected. Smart Slice currently assumes there is an active machine and will crash Cura when it tries to access a non-existing machine (None).

This is known to be an issue in the SmartSlicePropertyHandler constructor. We'll want to look for other cases of this. We'll also need to design and implement something to handle the situation appropriately since Smart Slice won't be usable until a machine is selected. What else might constitute an "invalid environment" that we should look out for?

PR #127, opened 2020-05-13
No printer bug fix

jake-kirkland-teton:
I moved the code that was causing the crash on first install, or if somehow all printers were removed. I changed it so now if the printer status changes, or on every run of Cura, it checks for a printer rather than just attempting to connect to it.

This fixes issue #85.


jake-kirkland-teton:
Alright, I renamed the function for clarity.

PR #132, opened 2020-05-15
Use a single search path for 3rd party modules

bradymadams:
This PR changes the dynamically updated search paths to only look in the top-level directory. I'm proposing this change to align with a new build structure that will allow us to build the plugin with specified branches of pywim and threemf, rather than using the hard coded versions in install-pywim.sh and install-3mf.sh.

@thopiekar, this approach appears to work fine, but I'm not sure if your original approach of having a separate search path for each 3rd party lib was put in for a reason. Let me know if this looks okay. This will allow me to more easily setup a build in Jenkins that will allow us to specify the specific branches of 3rd party libs for custom dev builds. This was inspired by the need to create a custom build from the cylinder select branches.


thopiekar:
@bradymadams: The intention was to be able to have prebuilt Python projects as submodules. Therefore, there was one directory per module within "cpython-common" and the other possible platform directories.
If you want to combine all install files into a platform's directory, it should be fine.

Regarding the scripts, I tested them and stripping works fine.

PR #135, opened 2020-05-22
Support modifier meshes in Smart Slice workflows

colman-hartinger:
Not ready to merge yet, still need to set up a signal listener on the message asking to remove modifier meshes for optimization otherwise smartslice will throw an error


colman-hartinger:
Everthing is working fine now and message will be displayed before optimization

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -189,9 +188,26 @@ def prepareJob(self, job_type, filename = None, filedir = None):
 
         # Checking whether count of models == 1
         mesh_nodes = getPrintableNodes()
+        mod_mesh = getModifierMeshes()
+
         if len(mesh_nodes) is not 1:
             Logger.log("d", "Found {} meshes!".format(["no", "too many"][len(mesh_nodes) > 1]))
             return None
+        if len(mod_mesh) is not 0:
```
bradymadams, line 196: Let's change this to `if len(mod_mesh) > 0:`. I realize the `not` was there before, but I never noticed it before. It's not very good practice because it compares the ids rather than the values. For low integers this works fine in Python, but it's probably best to just do regular integer comparisons.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -189,9 +188,26 @@ def prepareJob(self, job_type, filename = None, filedir = None):
...
+            placeholder_mesh = mod_mesh[0]
```
bradymadams, line 197: What happens if there's more than one mod mesh?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -189,9 +188,26 @@ def prepareJob(self, job_type, filename = None, filedir = None):
...
+            if mesh_nodes[0].hasChildren():
+                transformation = mesh_nodes[0].getWorldTransformation()
+                mesh_nodes[0].removeAllChildren()
+                op = GroupedOperation()
+                op.addOperation(AddSceneNodeOperation(placeholder_mesh,
+                                                      Application.getInstance().getController().getScene().getRoot()
+                                                      )
+                                )
```
bradymadams, line 205: There was some inconsistent styling that got introduced into our code early on, and this is one of them, so when you encounter this style of multi-line indentation lets fix it so that each new statement is only indented _one_ extra level, and the closing brace/parenthesis is aligned with it's matching opening statement. So in this case this should be:

```python
op.addOperation(
    AddSceneNodeOperation(
        placeholder_mesh,
        Application.getInstance().getController().getScene().getRoot()
    )
)
```

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -189,9 +188,26 @@ def prepareJob(self, job_type, filename = None, filedir = None):
...
+        mod_mesh = getModifierMeshes()
+
         if len(mesh_nodes) is not 1:
             Logger.log("d", "Found {} meshes!".format(["no", "too many"][len(mesh_nodes) > 1]))
             return None
+        if len(mod_mesh) is not 0:
+            placeholder_mesh = mod_mesh[0]
+            if mesh_nodes[0].hasChildren():
+                transformation = mesh_nodes[0].getWorldTransformation()
+                mesh_nodes[0].removeAllChildren()
```
bradymadams, line 200: Why are all of the children removed here? This seems risky to just remove all children without checking what they might be. If you are assuming that the mod mesh is a child of the printable node we should check for it explicitly and only remove that one child. There could be other children used for other purposes.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -786,6 +799,10 @@ def prepareOptimization(self):
         else:
             self.status = SmartSliceCloudStatus.Optimized
         self.updateSliceWidget()
+        propertyHandler = self.propertyHandler
+        if len(getModifierMeshes()) is not 0:
```
bradymadams, line 803: Integer comparison.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -786,6 +799,10 @@ def prepareOptimization(self):
...
+            propertyHandler.askToRemoveModMesh()
```
bradymadams, line 804: Is the temp variable, `propertyHandler` necessary? Can you just use `self.propertyHandler` here instead?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -189,9 +188,17 @@ def prepareJob(self, job_type, filename = None, filedir = None):
 
         # Checking whether count of models == 1
         mesh_nodes = getPrintableNodes()
+        mod_mesh = getModifierMeshes()
+
         if len(mesh_nodes) is not 1:
             Logger.log("d", "Found {} meshes!".format(["no", "too many"][len(mesh_nodes) > 1]))
             return None
+        if len(mod_mesh) > 0:
```
bradymadams, line 196: The length check isn't necessary anymore. The `for` loop will handle a zero length list appropriately.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -343,68 +350,82 @@ def _process_analysis_result(self, analysis : pywim.smartslice.result.Analysis,
...
+                    new_instance.resetState()  # Ensure that the state is not seen as a user state.
+                    settings.addInstance(new_instance)
+
+                op = GroupedOperation()
+                # First add node to the scene at the correct position/scale, before parenting, so the eraser mesh does not get scaled with the parent
+                op.addOperation(AddSceneNodeOperation(
+                    modifier_mesh_node,
+                    Application.getInstance().getController().getScene().getRoot()
+                    )
+                )
```
bradymadams, line 405: `AddSceneOperation` should be on it's own line, and the arguments should be indented an extra level.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -343,68 +350,82 @@ def _process_analysis_result(self, analysis : pywim.smartslice.result.Analysis,
...
+                                                    our_only_node)
+                                )
+                op.push()
+
+                # set modifier mesh node in root scene instead of as child of printable node
+                operation = GroupedOperation()
+                modifier_mesh_transform_matrix = Matrix(modifier_mesh.transform)
+                modifier_mesh_node.setTransformation(modifier_mesh_transform_matrix)
+                transformation = modifier_mesh_node.getWorldTransformation()
+                temp_node = modifier_mesh_node
```
bradymadams, line 416: I don't think `temp_node` is necessary. You should be able to just use `modifier_mesh_node`.

bradymadams:
@colman-hartinger the code is still adding the mod mesh twice. See the question in my last comment. Is that intended?

Also looks like the modifier mesh transformation is being set twice? Is that necessary or is it redundant?

colman-hartinger:
Ya the node needs to be added twice since when `removeChild()` is called it removes the node from the scene so I called `addOperation() ` to add back the now unconnected scene to the root node although there may be a better way to do that. Also the transformation has to be called twice because the transformation smart slice is getting from pywim is in relation to the parent part so you need to set that transformation to get the correct world transformation and then apply that to the mesh which is now operating in the root scene

bradymadams:
@colman-hartinger 

1. Why is the node added, then removed, and then added again? Can't we just add it once to the root?

2. Yeah, I'd like to see if there's a better way to handle that. This way seems like it could be a little tricky to maintain.

dougkenik:
So the workflow is not really working as expected right now. Here is an issue I encountered with the latest build:

Setup is below:
image

image

If you validate, and then optimize, it will spit back  modifier mesh with an infill density of 20% (see below)

image

If I change the infill density to 30%, I would expect Smart Slice to issue a warning about losing optimization results, and then moving back to Validate if the user accepts, or changing the density back to 20% if the user does not accept. Instead, it just thinks the part is optimized, and the SmartSlice button stays in "Preview" mode

colman-hartinger:
@dougkenik I'll setup a signal in the property handler to issue a warning about loss of results when the infill mesh density is changed

dougkenik:
@colman-hartinger  - this will need to happen whenever **any** mesh property is changed from an optimization.

colman-hartinger:
@dougkenik got it

dougkenik:
let's just make sure we are checking that (a) the modifier mesh is from the optimization algorithm, (b) the status is "optimized"

colman-hartinger:
ya I was thinking I would just change the sceneChanged function a bit to capture the warning, looks like it will issue it if anything in the printable mesh is changed

dougkenik:
Looking better. A few things I'm noticing:

1. If I bring a part in with modifier meshes, then validate. I get the below popup after validation. IT shouldn't display until I select "Optimize"
image
2. If I select "Cancel" from the popup in 1, the button stays on "Preview / Optimize" - it should change back to "Validate"
3. For some reason I'm allowed to select "Optimize" when the popup is present - which causes an error. We should force the user to take action in the popup and disable the "Optimize" button until they have allowed us to remove the modifier meshes or canceled
4. I optimized once and got some modifier meshes back. I then changed a parameter in the mod mesh and went back to "Validate." I validated, then Optimized again. Once the optimization was done, I received the "Lose Optimization Results" dialogue. This shouldn't happen, as the optimization just finished, and I changed nothing
4. Let's change the wording in the below popup to:
_Modifying this setting will invalidate your Smart Slice results.
Do you want to continue and lose your optimized results?_
image

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -184,25 +184,46 @@ def changed(self) -> bool:
...
+            infill = stack.getProperty(self._check[0], 'value')
+            thickness = stack.getProperty(self._check[1], 'value')
+            wThickness = stack.getProperty(self._check[2], 'value')
+            mesh = n.getMeshData()
+            if n.getName() == 'SmartSliceMeshModifier' and self.optimized == 0:
+                self._node = n
+                self._properties = (infill, thickness, wThickness)
+                self._meshData = mesh
+                self.optimized = 1
+            return n, (infill, thickness, wThickness), mesh
```
bradymadams, line 209: This logic is a little strange to me. Are you sure you always want it to return the values from the first iteration through the for loop?

Also, rather than using three individual variables for the properties, build a list by looping through the `_check` list. This will be more flexible if properties are ever added/removed on `_check`.

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -184,25 +184,46 @@ def changed(self) -> bool:
...
-            scene_root = CuraApplication.getInstance().getController().getScene().getRoot()
-            scene_root.addChild(self._node)
+            node = self._node.callDecoration("getStack").getTop()
+            self._node.setMeshData(self._meshData)
+            for name, value in zip(self._check, self._properties):
+                node.setProperty(name, 'value', value)
 
     def changed(self) -> bool:
-        return not (self._node is self.value())
+        # node = self.value()
```
bradymadams, line 223: remove commented out code.

SmartSlicePlugin/SmartSliceProperty.py
colman-hartinger: True Changed things to run inside the if statement

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -184,25 +184,46 @@ def changed(self) -> bool:
...
     def __init__(self):
         self._node = None
+        self._properties = None
+        self._meshData = None
+        self._check = [
+            'infill_sparse_density',
+            'top_bottom_thickness',
+            'wall_thickness'
+        ]
+        self.optimized = 0
```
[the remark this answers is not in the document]
colman-hartinger: removed this part when I made the changes that Doug was talking about

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -771,46 +805,51 @@ def prepareValidation(self):
...
         self.updateSliceWidget()
 
     def doOptimization(self):
+        if len(getModifierMeshes()) > 0:
+            self.propertyHandler.askToRemoveModMesh()
+        else:
+            self.optimization()
+
+    def optimization(self):
+        ModifierMesh.optimized = 0
```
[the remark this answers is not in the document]
colman-hartinger: switched to using a single function, not sure why I Split it up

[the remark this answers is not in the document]
colman-hartinger: got rid of this in my current commit

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -342,84 +365,94 @@ def _process_analysis_result(self, analysis : pywim.smartslice.result.Analysis,
...
+                    )
+                )
+                op.push()
+
+                # set modifier mesh node in root scene instead of as child of printable node
+                modifier_mesh_transform_matrix = Matrix(modifier_mesh.transform)
+                modifier_mesh_node.setTransformation(modifier_mesh_transform_matrix)
+                transformation = modifier_mesh_node.getWorldTransformation()
+                modifier_mesh_node.setParent(Application.getInstance().getController().getScene().getRoot())
+                modifier_mesh_node.setTransformation(transformation)
```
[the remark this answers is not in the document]
colman-hartinger: It seemed when setParent() is called it removed the transformation, I'll try  again without setting the global transformation once the new API is consumed

colman-hartinger:
@bradymadams for the multiple lines styling I guess I let my ide reformat the file to pep 8 styling should I keep all the other formatting changes and just fix the multi line statements?

bradymadams:
> @bradymadams for the multiple lines styling I guess I let my ide reformat the file to pep 8 styling should I keep all the other formatting changes and just fix the multi line statements?

@colman-hartinger I didn't notice anything else that needs reverted. It looks like it adjusted the argument type annotations to remove the whitespace after the argument name:

```
def foo(a: int)
```

But, I'm fine with that as that agrees with the Google python style guide.

For further clarification on the multi-line stuff, [here's the entry in the google style guide](http://google.github.io/styleguide/pyguide.html#34-indentation).

They indicate either approach is fine, but I really don't like the indentation aligned with the opening bracket because it defeats the purpose of multi-line statements, where you want to bring the text back to the left because of a long module/function name. And it results in more maintenance when module and function names inevitably change.

PR #150, opened 2020-06-15
R20.1 new pywim api updates

### LINKED ISSUE #104: Consume new pywim Smart Slice API

jake-kirkland-teton:
This implements the new changes to the Thor API and config/token files. We are now building config files to store the connection information rather than storing it within Cura preferences, and we are also now storing the user's login token in a file as well. The waiting loop for a smartslice job has been reworked to use the new API method.

Closes #104


SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -151,6 +159,64 @@ def job_status(self, value):
             self._job_status = value
             Logger.log("d", "Status changed: {}".format(self.job_status))
 
+    # This method reads the config file, and if either the results are malformed or the file does not exist, it calls to create one.
+    def getConfig(self):
+        filepath = os.path.join(str(Resources.getPath(12)), "SmartSlicePlugin\\SmartSliceConfig.cfg")
```
bradymadams, line 164: 1. Use `Resources.Plugins` instead of a magic number. It makes it more obvious what you're retrieving, and will also be future proofed against any changes to those values.
2. `getPath` already returns a string, so the call to `str` isn't necessary
3. Be careful with hard coding path separators. They are different on Windows and Linux. This call would create a file with a backslash character in it on Linux.
4. I might be missing something, but I think there could be a cleaner way to get the plugin directory. See here: https://github.com/tetonsim/is-cura-ui/blob/877877660e5e4998fd6d83353f2a0e45c4c201d5/SmartSlicePlugin/SmartSliceCloudConnector.py#L615

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -151,6 +159,64 @@ def job_status(self, value):
...
+            fileDict = json.load(open(filepath))
+            if fileDict.get("protocol") in ("https", "http") and fileDict.get("hostname") is not None and fileDict.get("port") is not None:
+                return fileDict
+            else:
+                raise IOError
+        except:
+            Logger.log("e", "Failure to read Config JSON: Rebuilding")
+            self.createConfigFile(filepath)
+            fileDict = json.load(open(filepath))
+            return fileDict
```
bradymadams, line 178: It's a pretty uncommon and hard to read style to throw exceptions just to divert the logic flow. Let's remove that and just check for the conditions that require the config file to be created.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -151,6 +159,64 @@ def job_status(self, value):
...
+            self.createConfigFile(filepath)
+
+        try:
+            fileDict = json.load(open(filepath))
+            if fileDict.get("protocol") in ("https", "http") and fileDict.get("hostname") is not None and fileDict.get("port") is not None:
+                return fileDict
+            else:
+                raise IOError
+        except:
+            Logger.log("e", "Failure to read Config JSON: Rebuilding")
```
bradymadams, line 175: Creating the config won't be an uncommon event, right? If not, let's change the log to a debug, and remove the word "failure". Errors indicate something unexpected happened, and I don't think this is an unexpected event.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -151,6 +159,64 @@ def job_status(self, value):
...
+
+    # If the config file is corrupt or has incorrect data or does not exist, one with default settings is created here.
+    def createConfigFile(self, filepath):
+        defaultClient = {"protocol" : "https", "hostname" : "test.smartslice.xyz", "port" : "443"}
+        json.dump(defaultClient, open(filepath,"w"))
+
+    # If our user has logged in before, their login token will be in the file.
+    def getToken(self):
+        #TODO: If no token file, try to login and create one. For now, we will just create a token file.
+        filepath = os.path.join(str(Resources.getPath(12)), "SmartSlicePlugin\\Token")
```
bradymadams, line 188: Rename the file to `.token`. It's common to prefix file names with a period when you don't want a user commonly getting into them.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -212,41 +278,41 @@ def processCloudJob(self, filepath):
...
 
-                Logger.log("e", "An error occured while sending and receiving cloud job: {}".format(task.error))
+                Logger.log("e", "An error occured while sending and receiving cloud job: {}".format(task.errors))
                 self.connector.propertyHandler._cancelChanges = False
                 return None
-            elif task.status == pywim.http.thor.TaskStatus.finished:
+            elif task.status == pywim.http.thor.JobInfo.Status.finished:
                 # Get the task again, but this time with the results included
-                task = self._client.result.get(id=task.id)
+                statusCode, task = self._client.smartslice_job(task.id, True)
```
bradymadams, line 310: smartslice_job_wait already returns the JobInfo with the results, so this call isn't necessary.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -151,6 +159,64 @@ def job_status(self, value):
...
+                token = json.load(open(filepath))
+                if token == "" or token is None:
+                    raise IOError
+                else:
+                    return token
+            except:
+                Logger.log("e", "Failure to read Token JSON: Rebuilding")
+                self.createTokenFile(filepath)
+                token = json.load(open(filepath))
+                return token
```
bradymadams, line 202: Same here about raising and catching exceptions to divert the logic. Let's rework this.

bradymadams:
There is also a conflict with the r20.1 branch that will need to be resolved before this can get merged. See below, it should show "This branch has conflicts...".

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: That is a much more effective way to get the path. Thanks for the tip!

Fair, I didn't think about hardcoding the separator. No longer an issue with the better way of getting the path.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: That is a fair point. I added them here since we need the try except to catch JSON loading errors or they cause Cura to fatally crash. I put the data checks outside the try/except now.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: This was specifically if the data in the config file was malformed, or if the JSON was unable to be loaded, causing an error to be thrown either by me from missing data, or from the JSON library. I changed it to a debug message.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Gotcha. It is done, thanks for the info, I actually didn't know that's why some files started with the dot.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Understood! Fixed.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Reworked the same way as above.

jake-kirkland-teton:
I tried to improve my variable names. Noted.

I moved the connection information to it's own config file because I interpreted the comment at the top of the thor-client-test.py to mean that you wanted a local file for the config information to make it easier to test certain things. I can certainly change it back to preferences if that is preferred. 

My apologies. I have added spaces.

Sorry about the conflicts. I thought I checked the correct spot for those in GitHub Desktop but I was incorrect. I will get those merged.

bradymadams:
> I moved the connection information to it's own config file because I interpreted the comment at the top of the thor-client-test.py to mean that you wanted a local file for the config information to make it easier to test certain things. I can certainly change it back to preferences if that is preferred.

cura.cfg is local, so I think it fits the need. I might have said something else contradictory, but I think the cura preferences is a perfectly acceptable place to put the connection parameters. I do agree, with using a separate file for the token, however, because of the risk of someone sharing a cura.cfg with another person.
 
> My apologies. I have added spaces.
> Sorry about the conflicts.

No apologies are necessary. This is the point of PR reviews. And conflicts are a natural result of multiple developers working in the same repo.

SmartSlicePlugin/SmartSliceCloudConnector.py
bradymadams: The logic is still a little tough to follow and there's some duplication that I think can be condensed. Some things to consider:

1. Can `createTokenFile` return the token, so you don't need to load the token after calling `createTokenFile`?

```python
token = self.createTokenFile(tokenFilePath)
```

2. There's no need to split the last if into a separate return. This is equivalent:

```python
if token == "" or token is None:
    token = self.createTokenFile(tokenFilePath)
return token
```

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -101,6 +108,44 @@ def job_status(self, value):
...
+        #TODO: Get the token from the login system correctly.
+        myToken = "[Insert Your Token Here]"
+        json.dump(myToken, open(tokenFilepath, "w"))
+
+    def fileExists(self, toCheckFilepath):
+        pluginDir = PluginRegistry.getInstance().getPluginPath("SmartSlicePlugin")
+        if not os.path.exists(pluginDir):
+            os.mkdir(pluginDir, 755)
+            return False
+        elif os.path.exists(pluginDir) and not os.path.exists(toCheckFilepath):
```
bradymadams, line 144: This check here is a bit disconnected. It assumes `toCheckFilePath` is a file in `pluginDir`, but it would be completely possible to call `fileExists` with a different file path.

Since we're getting the plug-in path, which will always exists, I think you can just remove the check for existence of the `pluginDir` altogether and the logic to create it. This would then remove the need for this function altogether because it will be just as easy to call `os.path.exists` where you're currently calling `fileExists`.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -27,6 +27,7 @@
 from UM.Scene.SceneNode import SceneNode
 from UM.Settings.SettingInstance import SettingInstance
 from UM.Signal import Signal
+from UM.Resources import Resources
```
bradymadams, line 30: I don't think this import is necessary anymore.

jake-kirkland-teton: It was not.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: You are absolutely correct, this is not needed.

jake-kirkland-teton:
I have the file opens now correctly encapsulated in with statements. 

Interesting on the naming conventions. I will make sure to adhere to that.

And the create token file method now returns the token.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -101,6 +107,33 @@ def job_status(self, value):
...
+            if token == "" or token is None:
+                token = self._createTokenFile(token_file_path)
+            return token
+
+    # If there is no token in the file, or the file does not exist, we create one.
+    def _createTokenFile(self, token_file_path):
+        #TODO: Get the token from the login system correctly.
+        my_token = "[Insert Your Token Here]"
+        with open(token_file_path, "r") as token_file:
+            json.dump(my_token, token_file, "w")
```
bradymadams, line 134: Shouldn't the file be opened for writing?

What is the `"w"` for in the call to `json.dump`? The documentation indicates that `dump` only takes 2 positional arguments. I think this would throw an exception.

Please make sure you test your changes.

image

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -101,6 +107,33 @@ def job_status(self, value):
             self._job_status = value
             Logger.log("d", "Status changed: {}".format(self.job_status))
 
+    # If our user has logged in before, their login token will be in the file.
+    def _getToken(self):
+        #TODO: If no token file, try to login and create one. For now, we will just create a token file.
+        token_file_path = os.path.join(PluginRegistry.getInstance().getPluginPath("SmartSlicePlugin"), ".token")
+        if not os.path.exists(token_file_path):
+            self._createTokenFile(token_file_path)
```
bradymadams, line 115: Should we capture the returned token and return it from this function? I would remove one level of indent on the return on line 127. Assign the return from _createTokenFile to token, and remove all other returns. That way you know this function will always return a token. I might be missing something, but it looks like this function will return None if the token file has to be created.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Yup. You are exactly right. I appear to have lost my mind for a second. I had a weird thing happen to my plugin files and ended up getting some bizarre errors and got distracted fixing them and didn't have it recreate the token file. That is on me.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Yeah I should have been doing that. Fixed. And the indent change is a good idea. I think you were dead on, it would have returned None. I hope I have it all fixed now. Thanks for catching that for me. I should have noticed that one.

bradymadams:
For reference on my second point see our style guide. The indentation of multi-line statements is discussed there: https://tetoncomposites.atlassian.net/wiki/spaces/DEV/pages/357040133/Coding+Style#Python

bradymadams:
Sorry, I also meant to link this in regards to the first point about using the `is` operator. Trying to do too much at once this morning...

https://stackoverflow.com/questions/306313/is-operator-behaves-unexpectedly-with-integers

PR #152, opened 2020-06-19
R20.1 save project

dougkenik:
Tapped into Cura's save functionality for projects and now we will save the job setup and any results when Cura saves a project. I checked that the data is being written correctly to the Cura 3MF, and it looks good.

I also restructured the code a bit to remove a few circular dependencies, and tried to clean up other places where the logic gets confusing.


colman-hartinger:
When I load a part into smart slice and add loading, anchors and max deflection and then try and save the results smart slice gives error saying `'NoneType' object has no attribute 'getResult'` and doesnt save anythinng

saveError

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -134,6 +138,7 @@ def _createTokenFile(self, token_file_path):
         return my_token
 
     def _handleThorErrors(self, http_error_code, returned_object):
+        self.setError(returned_object.exception)
```
colman-hartinger, line 141: This caused an exception during the validation step saying ''NoneType' object has no attribute 'exception''

dougkenik: Should be fixed

dougkenik:
Bugs are fixed. Give it another go at your convenience

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -540,9 +402,7 @@ def _onSaveDebugPackage(self, messageId: str, actionId: str) -> None:
         jobname = Application.getInstance().getPrintInformation().jobName
         debug_filename = "{}_smartslice.3mf".format(jobname)
         debug_filedir = self.app_preferences.getValue(self.debug_save_smartslice_package_location)
-        dummy_job = dummy_job.prepareJob(dummy_job.job_type,
-                                         filename= debug_filename,
-                                         filedir= debug_filedir)
+        dummy_job = dummy_job.prepareJob(filename= debug_filename, filedir= debug_filedir)
```
bradymadams, line 405: Formatting. Additional space after `=` in function call.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -728,27 +584,170 @@ def _onApplicationActivityChanged(self):
...
+            error = str(self._jobs[self._current_job].getError())
+            self.cancelCurrentJob()
+            Logger.logException("e", error)
+            Message(
+                title='Smart Slice job unexpectedly failed',
+                text=error
+            ).show()
+            return
+
+        self.propertyHandler._propertiesChanged = []
```
bradymadams, line 601: Without knowing, myself, the details of where and how `_propertiesChanged` is used, I'd ask if this is really what you want. If there are other references to this list somewhere this won't affect those, and the old list will be retained. `clear` would be better.

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -55,13 +69,45 @@ def _createQmlDialog(self, dialog_qml, directory = None, vars = None):
...
+        return about
+
+    def _saveState(self, output_object=None):
+        plugin_info = self._getMetadata()
+
+        # Build the Smart Slice job. We want to always build in case something has changed
+        job = self.cloud.smartSliceJobHandle.buildJobFor3mf()
+
+        cloudJob = self.cloud.cloudJob
+        if (cloudJob):
```
bradymadams, line 86: Formatting - remove the parens.

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -55,13 +69,45 @@ def _createQmlDialog(self, dialog_qml, directory = None, vars = None):
...
+            job.type = cloudJob.job_type
+
+        # Place the job in the metadata under our plugin ID
+        self._storage.setEntryToStore(plugin_id = plugin_info['id'], key = 'job', data = job.to_dict())
```
bradymadams, line 90: Formatting. No spaces around the `=` in this case (below calls as well).

http://google.github.io/styleguide/pyguide.html#36-whitespace

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -55,13 +69,45 @@ def _createQmlDialog(self, dialog_qml, directory = None, vars = None):
...
+
+        # Need to do some checks to see if we've stored the results for the active job
+        if cloudJob and cloudJob.getResult() and not cloudJob.saved:
+            self._storage.setEntryToStore(plugin_id = plugin_info['id'], key = 'results', data = cloudJob.getResult().to_dict())
+            cloudJob.saved = True
+        elif job.type == pywim.smartslice.job.JobType.validation and (not cloudJob or not cloudJob.getResult()):
+            self._storage.setEntryToStore(plugin_id = plugin_info['id'], key = 'results', data = None)
+        else:
+            pass
```
bradymadams, line 99: I don't think this `else` is necessary.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -42,16 +42,16 @@ class SmartSliceJobHandler:
...
-        self._connector = connector
+    def __init__(self, handler: SmartSlicePropertyHandler):
         self._all_extruders_settings = None
+        self._propertyHandler = handler
 
-    # Builds and checks a smart slice job for errors. Returns the job
-    def checkJob(self) -> pywim.smartslice.job.Job:
+    # Builds and checks a smart slice job for errors based on current setup defined by the property handler
+    # Will return the job, and a dictionary of error keys and associated error resolutions
+    def checkJob(self, machine_name = "printer"):
```
bradymadams, line 51: Formatting.

Also, can we add the type annotation back and update it to reflect the change? Use the `Tuple` type hint from the typing module.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -164,36 +164,28 @@ def checkJob(self) -> pywim.smartslice.job.Job:
...
             error_dict[err.error()] = err.resolution()
 
-        self._connector._proxy.errors = error_dict
-
-        return job
+        return job, error_dict
 
     # Builds a complete smart slice job to be written to a 3MF
-    def buildJobFor3mf(self) -> pywim.smartslice.job.Job:
+    def buildJobFor3mf(self, machine_name = "printer") -> pywim.smartslice.job.Job:
```
bradymadams, line 180: Formatting.

SmartSlicePlugin/SmartSliceCloudConnector.py
dougkenik: this is legacy. I'll try `clear`

dougkenik:
This should be good now. I branched off of this branch for the project open ticket, so let's wait until I get that merged back in here before we merge to r20.1

bradymadams:
> This should be good now. I branched off of this branch for the project open ticket, so let's wait until I get that merged back in here before we merge to r20.1

Can you open a separate PR for merging back into this branch? It will reduce the amount of diffs to review since this PR contains a large amount of diffs from the re-structuring that was done.

dougkenik:
> > This should be good now. I branched off of this branch for the project open ticket, so let's wait until I get that merged back in here before we merge to r20.1
> 
> Can you open a separate PR for merging back into this branch? It will reduce the amount of diffs to review since this PR contains a large amount of diffs from the re-structuring that was done.

Yes - still in the middle of the work, but I will create a new PR to merge the open work into this branch

bradymadams:
One thing I noticed while looking over the open project PR is that there are several changes to the structure of the data in the 3MF that we will need to account for in pywim, if we want to use this same structure when submitting a job to the Smart Slice engine.

1. The folder name has been changed from `SmartSlice` to `SmartSlicePlugin`
2. The file name is `plugin_metadata.json` and _all_ data goes in there. It doesn't look like we have control over the name, or can add other files.
3. Since all of the data is in one file, we have to combine the job, results, etc. We'll need to account for that when unpacking the contents of the file to get the job information.

dougkenik:
> One thing I noticed while looking over the open project PR is that there are several changes to the structure of the data in the 3MF that we will need to account for in pywim, if we want to use this same structure when submitting a job to the Smart Slice engine.
> 
> 1. The folder name has been changed from `SmartSlice` to `SmartSlicePlugin`
> 2. The file name is `plugin_metadata.json` and _all_ data goes in there. It doesn't look like we have control over the name, or can add other files.
> 3. Since all of the data is in one file, we have to combine the job, results, etc. We'll need to account for that when unpacking the contents of the file to get the job information.

This is all true, but I was thinking we could keep the data for saving and job submission separate for now. If you want to use the same structure, we can do that

bradymadams:
> This is all true, but I was thinking we could keep the data for saving and job submission separate for now. If you want to use the same structure, we can do that

I don't have a strong preference either way, but I do think we should at least update pywim to recognize the two different formats. It wouldn't take much work and it will allow us to easily grab 3MFs from Cura and troubleshoot them with our command line tools.

dougkenik:
> > This is all true, but I was thinking we could keep the data for saving and job submission separate for now. If you want to use the same structure, we can do that
> 
> I don't have a strong preference either way, but I do think we should at least update pywim to recognize the two different formats. It wouldn't take much work and it will allow us to easily grab 3MFs from Cura and troubleshoot them with our command line tools.

I'll create a branch in pywim and add the functionality

bradymadams:
> I'll create a branch in pywim and add the functionality

I think we can hold off for now and do it in a separate issue, if you want. I just wanted to mention it here so my thoughts didn't get lost.

dougkenik:
that's fine. I'll create a new issue

dougkenik:
> Overall it looks good to me, but I have a couple comments/requests:
> 
> 1. What do you think about storing the plugin version in the 3MF? We should probably have that in there so in future versions if we change the format we can check the version and see if we have to read it differently.
> 2. Do you think it would be advantageous to create a new class that you set to one key? I could go either way on this, but I could see it potentially getting cumbersome to manage saving/restoring multiple keys. Just a thought.
> 
> ```python
> class SomeNewClass(pywim.WimObject):
>     def __init__(self):
>         self.job = pywim.smartslice.job.Job()
>         etc...
> 
> x = SomeNewClass()
> 
> self._storage.setEntryToStore(plugin_id=plugin_info['id'], key='some_new_object', data=x.to_dict())
> ```

Yes, I'll add the version number to the stored data.

I am indifferent on storing everything as one class right now. It's super easy to manage as-is, and it allows us to update the job/result independently as data is created. I could see doing this in the future if it becomes bloated.

dougkenik:
In addition, we are storing all of the result data which comes back from the cloud job. The `Result `class has a `Meta `attribute which contains the metadata from the AWS job, and it is not populated right now. It would be good to have this data populated.

PR #154, opened 2020-06-23
R20.1 merge mod mesh flow

colman-hartinger:


SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -351,68 +384,75 @@ def _process_analysis_result(self, analysis : pywim.smartslice.result.Analysis,
...
+                op.push()
+
+                # Use the data from the SmartSlice engine to translate / rotate / scale the mod mesh
+                modifier_mesh_transform_matrix = Matrix(modifier_mesh.transform)
+                modifier_mesh_node.setTransformation(modifier_mesh_transform_matrix)
+                transformation = modifier_mesh_node.getWorldTransformation()
+                modifier_mesh_node.setParent(Application.getInstance().getController().getScene().getRoot())
+                modifier_mesh_node.setTransformation(transformation)
+                Application.getInstance().getController().getScene().sceneChanged.emit(modifier_mesh_node)
+                self.connector.propertyHandler.modMeshPropertyConnector(modifier_mesh_node)
```
colman-hartinger, line 456: @bradymadams removing the transformation here led to the modifier mesh not being on the build plate

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -197,29 +197,43 @@ def changed(self) -> bool:
...
-                return n
-        return None
+            stack = n.callDecoration("getStack").getTop()
+            properties = tuple([stack.getProperty(property, "value") for property in self._check])
+            if n.getName() == "SmartSliceMeshModifier":
+                n.setName("MeshModifier")
+                self._node = n
+                self._properties = properties
+            return n, properties
+        return None, None
```
colman-hartinger, line 219: @bradymadams not 100% sure about the logic here but when the return statement was inside the if statement the popup would consistently show after optimization

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -351,68 +384,75 @@ def _process_analysis_result(self, analysis : pywim.smartslice.result.Analysis,
...
+                modifier_mesh_node.addDecorator(BuildPlateDecorator(active_build_plate))
+                modifier_mesh_node.addDecorator(SliceableObjectDecorator())
+
+                stack = modifier_mesh_node.callDecoration("getStack")
+                settings = stack.getTop()
+
+                modifier_mesh_node_infill_pattern = SmartSliceJobHandler.INFILL_SMARTSLICE_CURA[modifier_mesh.print_config.infill.pattern]
+                definition_dict = {
+                    "infill_mesh": True,
+                    # "infill_pattern" : modifier_mesh_node_infill_pattern,
```
bradymadams, line 421: Why is the infill pattern commented out?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -351,68 +384,75 @@ def _process_analysis_result(self, analysis : pywim.smartslice.result.Analysis,
...
+                op.addOperation(AddSceneNodeOperation(
+                    modifier_mesh_node,
+                    Application.getInstance().getController().getScene().getRoot()
+                )
+                )
+                op.addOperation(SetParentOperation(
+                    modifier_mesh_node,
+                    our_only_node
+                )
+                )
```
bradymadams, line 445: Since the two open parens are on the same line put the two close parens on the same line. Same with lines 439-440.

SmartSlicePlugin/SmartSliceCloudConnector.py
bradymadams: That seems strange and we should probably understand what's going on here so we don't end up with an unexpected bug.

If you inspect `modifier_mesh_transform_matrix` and `transformation`:
1. are they the same or different before the setParent call?
2. are they the same or different after the setParent call?

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -122,6 +124,10 @@ def printerCheck(self):
     def hasModMesh(self) -> bool:
         return self._modifier_mesh.value() is not None
 
+    def modMeshPropertyConnector(self, mesh):
+        self._stack = mesh.callDecoration('getStack')
+        self._stack.propertyChanged.connect(self._onModMeshChanged)
```
bradymadams, line 129: Is this the only place `_stack` is used? If so, I don't see a reason that we need to introduce it as a class attribute - it could just be a local variable within this function, right?

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -63,6 +63,8 @@ def __init__(self, connector):
         self._selected_material = SmartSliceProperty.SelectedMaterial()
         self._scene = SmartSliceProperty.Scene()
         self._modifier_mesh = SmartSliceProperty.ModifierMesh()
+        self._stack = None
+        self._msg = None
```
bradymadams, line 67: Are these two new variables dedicated to modifier mesh usage? If so, let's use more descriptive names for class attributes since they have a much broader scope than local variables. It's important that other developers can easily identify what a class attribute is for.

SmartSlicePlugin/SmartSlicePropertyHandler.py
colman-hartinger: yup changed it back to a function variable

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: was going to ask you what to do with this. Trying to validate with the infill pattern saved into the 3mf causes an error with pywim saying invalid 3mf. So I wasn't sure the best way to go about handling it.

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: `modifier_mesh_transform_matrix` and `transformation` are constant throughout the setParent Call. I think using multiply with the printable node transformation and the local transformation for the mod mesh could be a bit cleaner of a solution though

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -188,9 +196,14 @@ def prepareJob(self, job_type, filename = None, filedir = None):
 
         # Checking whether count of models == 1
         mesh_nodes = getPrintableNodes()
+        mod_mesh = getModifierMeshes()
+
         if len(mesh_nodes) is not 1:
             Logger.log("d", "Found {} meshes!".format(["no", "too many"][len(mesh_nodes) > 1]))
             return None
+        for node in mod_mesh:
+            Logger.log("d", "Adding modifier mesh {} to validation")
```
bradymadams, line 205: The name needs to be passed to the log call.

SmartSlicePlugin/SmartSliceCloudConnector.py
bradymadams: Your changes look good. It reads much better now :+1:

SmartSlicePlugin/SmartSliceProperty.py
bradymadams: I'm still confused on the logic here. What's the point of changing the name? It seems risky - would the property tracking work for a second set of changes if the name has changed?

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -77,7 +77,7 @@ def checkJob(self) -> pywim.smartslice.job.Job:
         nodes = [normal_mesh] + getModifierMeshes()
 
         # Cycle through all of the meshes and se
-        for node in nodes:
+        for node in [normal_mesh]:
```
dougkenik, line 80: @colman-hartinger - why are you only checking these errors for the normal mesh? We need to check every mesh to make sure the extruder and material is set correctly, don't we?

SmartSlicePlugin/SmartSliceProperty.py
dougkenik: Agree with @bradymadams - this seems a little convoluted

SmartSlicePlugin/SmartSliceJobHandler.py
colman-hartinger: True I added another for loop to do an extruder check on the modifier meshes but I'm thinking It is safe to assume that if we are using the same extruder as the printable nodes than the material is set correctly. The reason I changed this is because adding the material of the modifier meshes to the job.bulk object was causing an error with the solvers

SmartSlicePlugin/SmartSliceProperty.py
colman-hartinger: Ya I've changed this to a bool check as Brady previously suggested.

SmartSlicePlugin/SmartSliceJobHandler.py
dougkenik: Where is the other check, and does it display the errors to the user in the error popup? The reason this routine is put into place is to dynamically tell the user when the setup is incorrect. If adding the materials to the job is causing an error, then we should investigate why

colman-hartinger: It's on my local branch, I'm waiting to test things out fully to push it up.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -351,68 +382,69 @@ def _process_analysis_result(self, analysis : pywim.smartslice.result.Analysis,
...
+                modifier_mesh_node.addDecorator(BuildPlateDecorator(active_build_plate))
+                modifier_mesh_node.addDecorator(SliceableObjectDecorator())
+
+                stack = modifier_mesh_node.callDecoration("getStack")
+                settings = stack.getTop()
+
+                modifier_mesh_node_infill_pattern = SmartSliceJobHandler.INFILL_SMARTSLICE_CURA[modifier_mesh.print_config.infill.pattern]
+                definition_dict = {
+                    "infill_mesh": True,
+                    # "infill_pattern": modifier_mesh_node_infill_pattern,
```
bradymadams, line 419: I think we can uncomment this now.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -233,8 +233,8 @@ def processCloudJob(self, filepath):
             self._handleThorErrors(thor_status_code, task)
             self.connector.cancelCurrentJob()
 
-        if task is not None:
-            Logger.log("d", "Job status after post'ing: {}".format(task.status))
+        # if task is not None:
```
dougkenik, line 236: Why was this commented out?

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -108,6 +105,22 @@ def checkJob(self) -> pywim.smartslice.job.Job:
...
+            job.chop.meshes.add(mesh)
+
+            # Check the active extruder
+            any_individual_extruder = all(map(lambda k : (int(active_extruder.getProperty(k, "value")) <= 0), SmartSlicePropertyHandler.EXTRUDER_KEYS))
+            if not ( int(active_extruder.getMetaDataEntry("position")) == 0 and int(emActive) == 0 and any_individual_extruder ):
+                errors.append(pywim.smartslice.val.InvalidSetup(
+                    "Invalid extruder selected for <i>{}</i>".format(node.getName()),
+                    "Change active extruder to Extruder 1"
+                ))
+
```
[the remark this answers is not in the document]
dougkenik: So, you're not checking the material for the modifier meshes? I agree that if all meshes must have the same extruder, only 1 material check is needed.... Maybe the material check gets pulled out of the for loop, and we consolidate the for loop for all meshes, then check the material for the normal mesh?

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -303,9 +307,10 @@ def askToRemoveModMesh(self):
             self.removeModMeshes(self._mod_mesh_removal_msg, 'continue')
         else:
             self._mod_mesh_removal_msg = Message(
-                title="",
+                title="Smart Slice Optimization",
```
dougkenik, line 310: I think we need to standardize the "Titles" of all these message boxes to "Smart Slice Info", "Smart Slice Warning" or "Smart Slice Error"

SmartSlicePlugin/SmartSliceProperty.py
colman-hartinger: Ya I think creating a `ModifierMesh` instance to check every modifier mesh added to the scene is a better aproach.

SmartSlicePlugin/SmartSliceJobHandler.py
colman-hartinger: from Dougs comment I consolidated this to a single loop for the printable and non printable nodes and moved the material check from the for loop to just check the printable node

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -714,14 +735,21 @@ def _process_analysis_result(self):
                 modifier_mesh_node,
                 Application.getInstance().getController().getScene().getRoot()
             ))
-            op.addOperation(SetParentOperation(modifier_mesh_node, our_only_node))
+            op.addOperation(SetParentOperation(
+                modifier_mesh_node,
+                Application.getInstance().getController().getScene().getRoot()
+                ))
```
bradymadams, line 741: Reduce indentation on closing parens by one level.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -73,10 +73,26 @@ def checkJob(self, machine_name="printer") -> Tuple[pywim.smartslice.job.Job, Di
...
 
-        # Get all of the normal and mod meshes
+        # Check the material
+        machine_extruder = getNodeActiveExtruder(normal_mesh)
+        guid = machine_extruder.material.getMetaData().get("GUID", "")
+        material = self._getMaterial(guid)
+
+        if not material:
+            errors.append(pywim.smartslice.val.InvalidSetup(
+                    "Material <i>{}</i> is not currently characterized for Smart Slice".format(machine_extruder.material.name),
```
bradymadams, line 83: Indentation.

SmartSlicePlugin/SmartSliceJobHandler.py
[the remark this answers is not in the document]
bradymadams: Do we need to ensure that the modifier meshes don't use a different extruder/material? Maybe that's already done somewhere?

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -62,7 +62,8 @@ def __init__(self, connector):
         self._extruder_properties = SmartSliceProperty.ExtruderProperty.CreateAll()
         self._selected_material = SmartSliceProperty.SelectedMaterial()
         self._scene = SmartSliceProperty.Scene()
-        self._modifier_mesh = SmartSliceProperty.ModifierMesh()
+        self._ModifierMesh = []
```
bradymadams, line 65: Naming convention doesn't match the other names in this class.

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -121,8 +122,27 @@ def printerCheck(self):
...
+    def loadModifierMesh(self, root):
+        for node in getModifierMeshes():
+            self._ModifierMesh.append(SmartSliceProperty.ModifierMesh(node))
+            self._properties += self._ModifierMesh
+            self.cacheChanges()
+            self.modMeshPropertyConnector(node)
+
+    def trackModifierMeshProperty(self, node):
+        self._ModifierMesh.append(SmartSliceProperty.ModifierMesh(node))
+        self._properties += self._ModifierMesh
```
bradymadams, line 138: Won't this re-add the entire list everytime resulting in duplicates?

image

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -168,7 +188,15 @@ def _onMaterialChanged(self):
...
-        self.confirmPendingChanges( [self._scene, self._modifier_mesh] )
+        self.confirmPendingChanges([self._scene])
+
+    def _onModMeshChanged(self, scene_node, infill_node):
+        for meshTracker in self._ModifierMesh:
+            self.confirmPendingChanges([meshTracker])
+
+    def _onModMeshMoved(self, infill_node):
+        for meshTracker in self._ModifierMesh:
+            self.confirmPendingChanges([meshTracker])
```
bradymadams, line 199: The brackets aren't necessary. `confirmPendingChanges` can take a single SmartSliceTrackedProperty object.

colman-hartinger:
@dougkenik it's probably slow because it's setting every property and transformation when a property is changed, I'll right some function to only set the property that was changed

SmartSlicePlugin/SmartSlicePropertyHandler.py
colman-hartinger: Every time a validation or optimization is run it deletes all the objects in the `_ModifierMesh` list and will add any modifier mesh in the scene  again

colman-hartinger: I did things this way because when I added an empty list to 'self._properties' it was deleted right away. Now I'm thinking it might be a better Idea to build some ModMesh object that contains all the tracked properties to have more efficient adding and deleting and do away with redundant code

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -121,8 +123,26 @@ def printerCheck(self):
...
+            if node.getName() not in list(self._modifier_mesh.keys()):
+                self.buildModifierMeshTrackedProperty(node)
+
+    def modMeshRemoved(self, node):
+        self._modifier_mesh = {}
+        for property in self._properties:
+            if type(SmartSliceProperty.ModifierMesh()) == type(property):
+                self._properties.remove(property)
+        for node in getModifierMeshes():
+            self.buildModifierMeshTrackedProperty(node)
```
bradymadams, line 145: When checking types `isinstance` is more robust and preferred.

I think we could make this more efficient.

```python
    def modMeshRemoved(self, node):
        for property in self._properties:
            if isinstance(property, SmartSliceProperty.ModifierMesh) and property._node is node:
                self._properties.remove(property)
                break
```

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -121,8 +123,26 @@ def printerCheck(self):
...
+        stack.propertyChanged.connect(self._onModMeshChanged)
+        node.parentChanged.connect(self.modMeshRemoved)
+
+    def loadModifierMesh(self, root):
+        for node in getModifierMeshes():
+            if node.getName() not in list(self._modifier_mesh.keys()):
+                self.buildModifierMeshTrackedProperty(node)
+
+    def modMeshRemoved(self, node):
+        self._modifier_mesh = {}
```
bradymadams, line 140: Be careful with re-assigning a dictionary like this. This creates a new dictionary and doesn't affect anything that might be referencing the original dictionary. `clear` is a better choice here.

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -168,7 +188,11 @@ def _onMaterialChanged(self):
         self.confirmPendingChanges(self._selected_material)
 
     def _onSceneChanged(self, changed_node):
-        self.confirmPendingChanges( [self._scene, self._modifier_mesh] )
+        self.confirmPendingChanges(self._scene)
+
+    def _onModMeshChanged(self, scene_node, infill_node):
+        for meshTracker in list(self._modifier_mesh.values()):
```
bradymadams, line 194: I don't think the call to `list` is necessary.

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -121,8 +123,26 @@ def printerCheck(self):
...
+        self._modifier_mesh[node.getName()] = SmartSliceProperty.ModifierMesh(node)
+        self._properties.append(self._modifier_mesh[node.getName()])
+        self.cacheChanges()
+        stack = node.callDecoration('getStack')
+        stack.propertyChanged.connect(self._onModMeshChanged)
+        node.parentChanged.connect(self.modMeshRemoved)
+
+    def loadModifierMesh(self, root):
+        for node in getModifierMeshes():
+            if node.getName() not in list(self._modifier_mesh.keys()):
```
bradymadams, line 136: I don't think the call to `list` is necessary.

SmartSlicePlugin/SmartSlicePropertyHandler.py
colman-hartinger: I think the break here could cause some redundancy because if there are multiple modifier meshes within the list and only one of them is removed then there could be duplicate objects tracking the same mod mesh

bradymadams: > I think the break here could cause some redundancy because if there are multiple modifier meshes within the list and only one of them is removed then there could be duplicate objects tracking the same mod mesh

Understood, however:
A. we should avoid duplicate entries, if possible.
B. If you don't `break`, and there is a duplicate entry you may end up with a bug because modifying a list that you are looping over can cause unexpected behavior. Try running this code and you'll see what I mean:

```python
x = [1, 2, 2, 4]
for i in x:
    print(i)
    if i == 2:
        x.remove(i)
print(x)
```

colman-hartinger: I guess I also did a bad job at naming since the node object passed to the function is the parent node from `        node.parentChanged.connect(self.modMeshRemoved)` so that check would never hit. I see what you mean about the break though I have run into that problem before

colman-hartinger:
@bradymadams I think it could be troublesome removing the _modifier_mesh dictionary since sometimes 
`loadModifierMesh` would be called multiple times on the same part which is why I used the dictionary, so duplicate instances wouldn't be added

SmartSlicePlugin/SmartSlicePropertyHandler.py
colman-hartinger: do you think I could create a signal that emits the modifier mesh node and connect the 'parentChagned' signal to that?

bradymadams: > do you think I could create a signal that emits the modifier mesh node and connect the `parentChanged` signal to that?

Not quite sure I follow what you mean by creating a new signal.

I would create a new lambda or function for a each mod mesh that contains the mod mesh scene node and connect that to parent changed. That way you don't need to search for anything. Let me know if you want more info/details.

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -121,8 +122,27 @@ def printerCheck(self):
             self._activeMachineManager.activeMachine.propertyChanged.connect(self._onGlobalPropertyChanged)
             self._activeMachineManager.activeMaterialChanged.connect(self._onMaterialChanged)
 
-    def hasModMesh(self) -> bool:
-        return self._modifier_mesh.value() is not None
+    def buildModifierMeshTrackedProperty(self, node):
+        self._properties.append(SmartSliceProperty.ModifierMesh(node, node.getName()))
+        Logger.log("d", "Tracking properties for {}".format(node.getName()))
+        self.cacheChanges()
```
bradymadams, line 128: Why do _all_ the properties need to be re-cached here?

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -121,8 +122,27 @@ def printerCheck(self):
...
+        stack = node.callDecoration('getStack')
+        stack.propertyChanged.connect(self._onModMeshChanged)
+        node.parentChanged.connect(self._properties[-1].parentChanged)
```
bradymadams, line 131: This could be problematic if signals are ever emitted in threads (which I can't say for sure doesn't happen), and the properties list was changed between lines 126 and 131. The fix is to assign the new `ModifierMesh` instance to a local variable, append it to properties, and then use that variable name here on line 131.

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -121,8 +122,27 @@ def printerCheck(self):
...
+
+    def loadModifierMesh(self, root):
+        names = [mesh.mesh_name for mesh in self._properties if isinstance(mesh, SmartSliceProperty.ModifierMesh)]
+        for node in getModifierMeshes():
+            if node.getName() not in names:
+                self.buildModifierMeshTrackedProperty(node)
+
+    def modMeshRemoved(self, parent_node):
+        for property in self._properties:
+            if isinstance(property, SmartSliceProperty.ModifierMesh) and property.parent_changed is True:
```
bradymadams, line 142: The `is True` can be omitted.

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -168,7 +188,12 @@ def _onMaterialChanged(self):
         self.confirmPendingChanges(self._selected_material)
 
     def _onSceneChanged(self, changed_node):
-        self.confirmPendingChanges( [self._scene, self._modifier_mesh] )
+        self.confirmPendingChanges(self._scene)
+
+    def _onModMeshChanged(self, scene_node, infill_node):
+        for meshTracker in self._properties:
+            if isinstance(meshTracker, SmartSliceProperty.ModifierMesh):
+                self.confirmPendingChanges(meshTracker)
```
bradymadams, line 196: It would be more efficient to send all of the ModifierMesh properties at once. Take a look at `_onSelectToolPropertyChanged` for an example. You can use `isinstance` and `_properties` in the filter call.

SmartSlicePlugin/SmartSlicePropertyHandler.py
colman-hartinger: when the part gets loaded in the properties aren't cached so this will initialize things for property tracking

colman-hartinger: I could right a separate function to just cache the mod mesh properties though

bradymadams: Can't you just call the `cache` function on the ModifierMesh object you just created?

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -126,14 +126,14 @@ def buildModifierMeshTrackedProperty(self, node):
         modMesh = SmartSliceProperty.ModifierMesh(node, node.getName())
         self._properties.append(modMesh)
         Logger.log("d", "Tracking properties for {}".format(node.getName()))
-        self.cacheChanges()
+        [p.cache() for p in self._properties if isinstance(p, SmartSliceProperty.ModifierMesh)]
```
bradymadams, line 129: Why can't you just call `modMesh.cache()` rather than calling cache on every existing property?

colman-hartinger: didn't even think about that

PR #162, opened 2020-07-06
R20.1 login logout dialog

### LINKED ISSUE #106: Login/logout workflow dialogs
Implement the various login workflow dialogs and messages as mocked up.

jake-kirkland-teton:
Trying out a draft PR. This isn't quite done yet, there are a few more changes to be made, noted in the commit message. Thought I would get it out there so you can take a look and make sure I'm on the right track.

Will close #106


SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -635,6 +642,51 @@ def updateColorUI(self):
...
         self.updateColorMaxDisplacement()
 
+    #
+    #    Login Dialog
+    #
+
+    loginUsernameChanged = pyqtSignal()
+    loginPasswordChanged = pyqtSignal()
+    loginResultChanged = pyqtSignal()
+    badCredentialsChanged = pyqtSignal()
```
bradymadams, line 652: Is there anything that connects to these signals? Are they necessary?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -218,19 +159,175 @@ def processCloudJob(self, filepath):
...
+
+class SmartSliceCloudVerificationJob(SmartSliceCloudJob):
+
+    def __init__(self, connector) -> None:
+        super().__init__(connector)
+
+        self.job_type = pywim.smartslice.job.JobType.validation
+
+
+class SmartSliceCloudOptimizeJob(SmartSliceCloudVerificationJob):
```
bradymadams, line 207: Should this inherit from `SmartSliceCloudJob`?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -218,19 +159,175 @@ def processCloudJob(self, filepath):
...
+    def __init__(self, connector):
+        super().__init__()
+        self._client = None
+        self.connector = connector
+        self.extension = connector.extension
+        self._token_file_path = os.path.join(PluginRegistry.getInstance().getPluginPath("SmartSlicePlugin"), ".token")
+        self._token = None
+
+        self._valid_token = None
+        self.login_success = None
```
bradymadams, line 224: `_token`, `_valid_token` and `login_success` seem redundant. Can we condense down into one attribute? Perhaps provide an `@property` function to yield if the token is valid or not for use outside of the class.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -274,54 +371,24 @@ def processCloudJob(self, filepath):
...
+            error_message.setText(i18n_catalog.i18nc("@info:status", "SmartSlice Server Error (400: Bad Request):\n{}".format(returned_object.error)))
+            error_message.show()
+        elif http_error_code == 401:
+            error_message = Message()
+            error_message.setTitle("Smart Slice Thor API")
+            error_message.setText(i18n_catalog.i18nc("@info:status", "SmartSlice Server Error (401: Unauthorized):\nAre you logged in?"))
+            error_message.show()
+        else:
+            error_message = Message()
+            error_message.setTitle("Smart Slice Thor API")
```
bradymadams, line 387: Let's drop the usage of "Thor" since it's an internal name.

There are several duplicate lines within each condition in this if statement that can be consolidated.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -373,6 +443,9 @@ def __init__(self, proxy: SmartSliceCloudProxy):
 
         self.saveSmartSliceJob = Signal()
 
+        self.loginResult = None
```
bradymadams, line 446: Is this attribute used anywhere? I see it is set in `onLoginButtonClicked`, but I don't see it being used elsewhere, but maybe I missed it.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -819,3 +895,10 @@ def onSecondaryButtonClicked(self):
                 self.cancelCurrentJob()
         else:
             Application.getInstance().getController().setActiveStage("PreviewStage")
+
+    def onLoginButtonClicked(self):
+        self.api_connection = SmartSliceAPIClient(self)
```
bradymadams, line 900: Unless I'm missing something, let's instantiate `api_connection` in the constructor. We should always setup attributes entirely in the constructor, unless we have no choice. It helps avoid bugs where an object is unexpectedly in an incomplete state.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -218,19 +159,175 @@ def processCloudJob(self, filepath):
...
+        else:
+            self.login_success = False
+
+        Logger.log("d", "SmartSlice HTTP Client: {}".format(self._client.address))
+
+    def _login(self):
+        username = self.connector._proxy.loginUsername
+        password = self.connector._proxy.loginPassword
+
+        Logger.log("d", username + " " + password)
```
bradymadams, line 265: Let's remove this. Writing the password in plain text to log files is a security risk.

SmartSlicePlugin/SmartSliceCloudProxy.py
jake-kirkland-teton: Fair warning, I haven't fully cleaned the code from things I had a use for and decided against. 

And I think, if I understand correctly, which I may not, the signals are needed for the login popup to receive information changes? loginResult could probably go, I'm not sure if we need that one anymore.

For Username and Password I intended to use them to store the data in the preferences for persistence. Is that the way to go?

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Good question. I hadn't looked at it or changed it actually. I will take a look.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: _token holds the actual token once it is obtained. I think we do need that so it can be set to the client object.

I might be able to condense valid token and login success. I will see what I can do about the @property.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Understood.

Good point. Fixed.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Absolutely. This was a personal logger statement I was using to check and I hadn't removed it yet. Got it taken care of.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: For some reason this causes an infinite loop. I intended to open the connection upon tab change eventually, but I will try to figure out why it loops weirdly.

SmartSlicePlugin/SmartSliceCloudConnector.py
bradymadams: But couldn't you provide a proxy for `valid_token` and `login_success`?

```python
@property
def logged_in(self):
    return self._token is not None
```

I realize _token could have an invalid token, but I don't see any reason to keep _token populated once we know it's no good.

SmartSlicePlugin/SmartSliceCloudProxy.py
bradymadams: > And I think, if I understand correctly, which I may not, the signals are needed for the login popup to receive information changes? loginResult could probably go, I'm not sure if we need that one anymore.

I don't think they are necessary for that. The `pyqtProperty` methods handle the exchange of data between Python and QML. You could try removing the `notify` argument on the `pyqtProperty` statements to check.

Signals are used to connect a function to get a callback when a particular event takes place. In this case these would be used if you wanted a function called every time the username text changes, for example, but I don't see any use cases for that.

> For Username and Password I intended to use them to store the data in the preferences for persistence. Is that the way to go?

I don't see where the signals would come into play there. However, we shouldn't store the password anywhere. That's a security risk.

jake-kirkland-teton: Ah gotcha. I don't think I understood the signal system correctly.

I will try that, I bet you are right.

Fair enough, I will just store the username in that case. Good point. Wasn't thinking.

jake-kirkland-teton:
Got most of the requested changes in place. All good stuff.

Since I moved the properties from the proxy side, badCredentials has ceased working and I'm having a hard time figuring out why. Any advice Brady?

I probably missed a few minor cleanup things as well. I'm getting tired.

bradymadams:
> Got most of the requested changes in place. All good stuff.
> 
> Since I moved the properties from the proxy side, badCredentials has ceased working and I'm having a hard time figuring out why. Any advice Brady?
> 
> I probably missed a few minor cleanup things as well. I'm getting tired.

I took a look and it was an easy enough fix that I just committed it, so see my last commit.

Also, I added the `constant` flag to the pyqtProperties to get rid of the QML warning about un-notifiable properties. [See this SO question for more details](https://stackoverflow.com/questions/6728615/warning-about-non-notifyable-properties-in-qml).

And then one other thing I noticed is you'll need to add back in the registration of the SmartSliceCloudProxy QML type. That is still used heavily in other QML elements.

SmartSlicePlugin/SmartSliceCloudLogin.qml
```diff
@@ -10,6 +11,7 @@ import SmartSlice 1.0 as SmartSlice
 UM.Dialog
 {
     title: catalog.i18nc("@title:window", "Cloud login")
+    closeOnAccept: false
```
dougkenik, line 14: We should get rid of the title bar and exit button on the top. Check out this command:

flags: ~Qt.WindowTitleHint | ~Qt.WindowCloseButtonHint;

SmartSlicePlugin/SmartSliceCloudLogin.qml
```diff
@@ -45,79 +47,163 @@ UM.Dialog
...
+
+        Text {
+            id: noAccountText
+
+            Layout.alignment: Qt.AlignCenter
+
+            font: UM.Theme.getFont("default")
+            color: UM.Theme.getColor("text")
+
+            text: "<a href='https://test.smartslice.xyz/static/account.html'><b>Don't have an account?</b></a>"
```
bradymadams, line 174: We may need to look into grabbing the hostname and/or URLs from a config. The production hostname will be different, and once we get our production account management pages up the URLs will be different.

bradymadams:
> 3. I think it might be best to break out the QML for the login into its own file and use it as a reusable component to cut down code in `SmartSliceMain.qml.` See the ResultTable branch for how to do it. @bradymadams - what do you think on this topic?

Yes, I did notice that file is getting rather large. Splitting it up would greatly improve maintainability.

> 4. Were you able to figure out how to disable the background if this rectangle is present? It's not critical, but would be nice

I noticed this as well. I think at the very least the Validate button should be disabled. I was able to move the part into view, apply loads and click Validate (the job obviously failed because there was no valid token).

dougkenik:
> > 1. I think it might be best to break out the QML for the login into its own file and use it as a reusable component to cut down code in `SmartSliceMain.qml.` See the ResultTable branch for how to do it. @bradymadams - what do you think on this topic?
> 
> Yes, I did notice that file is getting rather large. Splitting it up would greatly improve maintainability.
> 
> > 1. Were you able to figure out how to disable the background if this rectangle is present? It's not critical, but would be nice
> 
> I noticed this as well. I think at the very least the Validate button should be disabled. I was able to move the part into view, apply loads and click Validate (the job obviously failed because there was no valid token).

At the minimum we should make the toolbar (use cases / requirements) and the button on the bottom right invisible (inactive) until the user is signed in. That should alleviate any major issues

jake-kirkland-teton:
1. Alright sounds good. I will work on the styling.

2. Can do.

3. Alright. I will get it moved back into it's own file.

4. I looked at a couple ideas for this but I wasn't able to find anything too amazing. There might be a way to draw a transparent rectangle behind it that covers the entire stage and captures all the mouse events. I will try that way and see if it works. If not, I will make the toolbar/buttons inactive.

SmartSlicePlugin/stage/ui/SmartSliceMain.qml
```diff
@@ -16,6 +16,7 @@
 //  API Imports
 import QtQuick 2.7
 import QtQuick.Controls 2.2
+import QtQuick.Controls 1.4 as QQC1
```
bradymadams, line 19: No longer necessary.

jake-kirkland-teton:
Alright. Take a look again fellas. Thanks for the feedback. I got the password stuff fixed, and I think the popup should look good on everyone's system now. Let me know if your spacing is still off Brady.

jake-kirkland-teton:
Hmmm. OK. I thought I had accounted for a potential issue of the spacing on your setup Brady, but apparently that was not the case. I will try to get that adjusted. 

I should be able to capture the enter press now. It was a little weird with the way it was handled with the UM.Dialog box but we are not using that. Thanks for reminding me. 

Doug:
1) Interesting, I believe you are correct. It does look a little bit cut off, and I'm not sure why that would be. I'll look into it.

2) I think I can manage that. Some of the spacing and margins has been a little... finicky.

jake-kirkland-teton:
Alright. Changes made. Let me know how it looks.

PR #164, opened 2020-07-16
R20.1 multiple bcs cherry1

colman-hartinger:
Still some tiny things in the qml that I need to fix but all the functionality should be there


SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -91,193 +55,79 @@ def getInstance():
...
-
-        if step and len(step.boundary_conditions) > 0:
-            self.anchor_face.triangles = self._interactive_mesh.triangles_from_ids(step.boundary_conditions[0].face)
 
-        self.redraw()
-
-        return
+        LoadFaces().loadAnchorFace(step, smart_slice_node)
+        LoadFaces().loadLoadFace(step, smart_slice_node)
+        LoadFaces().setOrigin(normal_mesh)
```
bradymadams, line 84: This is very strange to me. You're instantiating a temporary object and then calling a function on it. That leads me to believe these functions are just acting as a proxy to something else, and thus are not necessary (cut out the middle man).

SmartSlicePlugin/select_tool/SmartSliceSelectHandle.py
```diff
@@ -1,231 +1,82 @@
...
 
-        area_2 = SmartSliceSelectHandle._threePointArea2(v1, v2, v3)
-        alpha = SmartSliceSelectHandle._threePointArea2(point, v2, v3) / area_2
-        beta = SmartSliceSelectHandle._threePointArea2(point, v3, v1) / area_2
-        gamma = SmartSliceSelectHandle._threePointArea2(point, v1, v2) / area_2
-
-        total = alpha + beta + gamma
-
-        return total > 0.99 and total < 1.01
+class LoadFaces:
```
bradymadams, line 18: This class is just essentially a collection of static functions acting as proxies to other classes. This tells me the functionality within here probably belongs elsewhere.

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -91,193 +55,79 @@ def getInstance():
...
-                    camTool = controller.getCameraTool()
-                    aabb = sn.getBoundingBox()
-                    if aabb:
-                        camTool.setOrigin(aabb.center)
+        return
 
     def _onSelectedFaceChanged(self, curr_sf=None):
-        if not self.getEnabled():
-            return
+        """"
```
bradymadams, line 96: extra quote

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -91,193 +55,79 @@ def getInstance():
...
+        gets face id and triangles from current face selection
```
bradymadams, line 97: Capitalize "gets"

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -33,23 +35,29 @@ def beginRendering(self):
...
 
+            uniforms = {}
+            overlay = False
+
+            if isinstance(node, SmartSliceScene.AnchorFace):
+                uniforms['diffuse_color'] = [1., 0.4, 0.4, 1.]
+                overlay = True
+            elif isinstance(node, SmartSliceScene.LoadFace):
+                uniforms['diffuse_color'] = [0.4, 0.4, 1., 1.]
+                overlay = True
```
bradymadams, line 46: I realize I wrote the code above, but this wasn't a good design. Let's retrieve the color values from a method in the AnchorFace and LoadFace classes so we don't have to check the instance types here.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -143,10 +165,9 @@ def onStageSelected(self):
         self._connector.updateSliceWidget()
 
     def extruderPropertyChanged(self, key: str, property_name: str):
-        if key in SmartSlicePropertyHandler.EXTRUDER_KEYS:
             self.extruderChanged()
-
-    # This creates the popup dialog that informs the user that only the
+    #
+    # # This creates the popup dialog that informs the user that only the
```
bradymadams, line 170: strange change to the comment format

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -115,6 +108,25 @@ def __init__(self, connector):
...
+
+    def faceAdded(self, face):
+        prop = SmartSliceProperty.SmartSliceFace(face)
+        prop.cache()
+        self._properties.append(prop)
+
+    def faceChanged(self):
+        face_list = [face for face in self._properties if isinstance(face, SmartSliceProperty.SmartSliceFace)]
+        for face in face_list:
+            self._onFaceChanged(face)
```
bradymadams, line 121: Is there a reason you don't just call confirmPendingChanges with the whole list?

```python
self.confirmPendingChanges(face_list)
```

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -0,0 +1,187 @@
...
+from ..utils import findChildSceneNode, getPrintableNodes
+from ..stage import SmartSliceScene
+
+
+class BoundaryConditionListModel(QAbstractListModel):
+    Anchor = 0
+    Force = 1
+
+    loadDirectionChanged = pyqtSignal(SmartSliceScene.LoadFace)
+    loadMagnitudeChanged = pyqtSignal(SmartSliceScene.LoadFace)
```
bradymadams, line 19: It doesn't look like these signals are used anywhere now. If not, let's just remove them.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -0,0 +1,394 @@
...
+
+    def removeFace(self, bc):
+        self.removeChild(bc)
+        self.faceRemoved.emit(bc)
+
+    def magnitudeChanged(self):
+        self.loadPropertyChanged.emit()
+
+    def addAnchorFace(self, step):
+        if step and len(step.boundary_conditions) > 0:
```
bradymadams, line 86: This check isn't necessary. Put a type annotation on the step argument to indicate the type and that it's not optional. The for loop won't execute if the list is empty.

Same with line 93.

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -91,193 +59,79 @@ def getInstance():
...
-
-        self.redraw()
-
-        return
 
-    def _calculateMesh(self):
-        scene = Application.getInstance().getController().getScene()
-        nodes = Selection.getAllSelectedObjects()
+        smart_slice_node.addAnchorFace(step)
+        smart_slice_node.addLoadFace(step)
```
bradymadams, line 87: `addAnchorFace` and `addLoadFace` are kind of misleading names. What you're actually doing is adding the step, but in two separate calls. I would merge the functionality of those two functions and rename it to `loadStep`.

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -310,41 +164,24 @@ def getSelectionMode(self):
...
-
     def getLoadSelectionActive(self):
         return self._selection_mode is SelectionMode.LoadMode
 
-    def defineSteps(self):
-
+    @staticmethod
+    def defineSteps():
+        """
+        adds all selected faces and associated data from loads/anchors to 'steps' to send to validation
```
bradymadams, line 179: Use appropriate capitalization in doc strings. However, this doc string is mis-leading. It's not adding anything, it's creating a list of steps and returning it.

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -310,41 +164,24 @@ def getSelectionMode(self):
...
-                self.load_face.triangles
-            )
-
     def getLoadSelectionActive(self):
         return self._selection_mode is SelectionMode.LoadMode
 
-    def defineSteps(self):
-
+    @staticmethod
+    def defineSteps():
```
bradymadams, line 177: Good job identifying that this is a static method, however, when you see a function that's static it might be an indicator that the functionality is misplaced. For example:

```python
class A:
    # A has functions and attributes

class B:
    @staticmethod
    def do_something():
        a = get_A()
        # do a bunch of stuff with a
```

This tells me `do_something` actually belongs in class A:

```python
class A:
    # A has functions and attributes

    def do_something(self):
        # do a bunch of stuff with self
```

In this case it looks like `defineSteps` should be moved to SmartSliceScene.Root. I'd also rename it to `createSteps` and add a return type annotation so it's clear what the call is for.

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -8,8 +8,11 @@
 from cura.BuildVolume import BuildVolume
 from cura.Scene.ConvexHullNode import ConvexHullNode
 
+from .stage import SmartSliceScene
```
bradymadams, line 11: Another added unused import. Please be careful about this.

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -0,0 +1,185 @@
...
+from ..utils import findChildSceneNode, getPrintableNodes
+from ..stage import SmartSliceScene
+
+
+class BoundaryConditionListModel(QAbstractListModel):
+    Anchor = 0
+    Force = 1
+
+    # loadDirectionChanged = pyqtSignal(SmartSliceScene.LoadFace)
+    # loadMagnitudeChanged = pyqtSignal(SmartSliceScene.LoadFace)
```
bradymadams: Don't comment out unused code, just remove it.

SmartSlicePlugin/select_tool/BoundaryConditionList.py
bradymadams: FYI, I added these signals back in. My request to remove them was wrong. They are needed for the related pyqtProperties. They are unused, but Qt still wants them, otherwise it complains.

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -0,0 +1,185 @@
...
+
+        self._bcs = []  # List[HighlightFace]
+        self._bc_type = BoundaryConditionListModel.Anchor
+        self._smart_slice_scene_node = None
+        self._active_node = None
+
+    def _reset(self, *kwargs):
+        """
+        when parent node is changed and printable nodes is 0 remove all loads and anchors
+        """
```
bradymadams, line 36: Capitalization.

> When parent node is changed

It doesn't look like anything about a parent node is checked. Your doc string should refer to what the function _does_, not what events occur outside of the function when it's called.

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -0,0 +1,185 @@
...
+        if len(getPrintableNodes()) is 0:
```
bradymadams, line 37: Use `==` when comparing integers. `is` works in this case because the RHS is a number less than 256, but this isn't a good habit, because using `is` on integers will eventually come back to bite you as a bug.

https://stackoverflow.com/questions/306313/is-operator-behaves-unexpectedly-with-integers

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -0,0 +1,406 @@
...
+    transformation = Matrix()
+    transformation.setRow(1, [0, 0, 1, 0])
+    transformation.setRow(2, [0, -1, 0, 0])
+    _, rotation, _, _ = transformation.decompose()
+
+    def __init__(self):
+        super().__init__(name='_SmartSlice', visible=True)
+
+        self._interactive_mesh = None
+        self._aabb = None
```
bradymadams, line 60: I don't think this is necessary as a class attribute. You can grab this from `self.getParent().getBoundBox()` when you need it.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -0,0 +1,397 @@
...
+    def initialize(self, parent: SceneNode):
+        parent.addChild(self)
+
+        mesh_data = parent.getMeshData()
+
+        if mesh_data:
+            Logger.log('d', 'Compute interactive mesh from SceneNode {}'.format(parent.getName()))
+
+            self._interactive_mesh = makeInteractiveMesh(mesh_data)
+            self._aabb = parent.getBoundingBox()
```
bradymadams, line 63: `_aabb` isn't used anymore, right?

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -0,0 +1,397 @@
...
+
+class Root(SceneNode):
+    faceAdded = Signal()
+    faceRemoved = Signal()
+    loadPropertyChanged = Signal()
+
+    transformation = Matrix()
+    transformation.setRow(1, [0, 0, 1, 0])
+    transformation.setRow(2, [0, -1, 0, 0])
+    _, rotation, _, _ = transformation.decompose()
```
bradymadams, line 49: This is kind of a strange place to make calls to functions. Is this necessary at the class level? I think this might be cleaner in a static method.

bradymadams:
@colman-hartinger this is looking good to me now. Is there anything else you have left to do? I noticed some TODO comments in the code.

Can you also merge `r20.1` when you get a chance?

colman-hartinger:
@bradymadams just megred r20.1 and the only relevant TODO note I saw was something you wrote about getting rid of self._bcs and dynamically getting the boundary condition nodes. I can look into that if need be otherwise this branch is good to merge

colman-hartinger:
wait hold on a sec realized I didn't merge everything properly

bradymadams:
Closing. See #166.

PR #168, opened 2020-07-22
R20.1 table

dougkenik:
Multiple results from optimization now come back, and are displayed in a table. The user can select any row, and all of the data in the popup, slice properties, and scene update automatically. User can sort by any column, leave and come back to the stage with table state remembered, and preview directly from the table.  Can save the state of the selected row in the table, which is repopulated in project open.


SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -371,7 +359,10 @@ def submitSmartSliceJob(self, cloud_job, threemf_data):
             self.connector.cancelCurrentJob()
 
         if task is not None:
-            Logger.log("d", "Job status after post'ing: {}".format(task.status))
+            try:
+                Logger.log("d", "Job status after posting: {}".format(task.status))
+            except:
+                pass
```
colman-hartinger, line 365: this could possibly be cleaner using `hasattr(task, "status")`

SmartSlicePlugin/SmartSliceCloudStatus.py
```diff
@@ -0,0 +1,27 @@
...
+        return (
+            SmartSliceCloudStatus.Underdimensioned,
+            SmartSliceCloudStatus.Overdimensioned,
+            SmartSliceCloudStatus.RemoveModMesh
+        )
+
+    @staticmethod
+    def busy():
+        return (SmartSliceCloudStatus.BusyValidating, SmartSliceCloudStatus.BusyOptimizing)
+
```
colman-hartinger, line 27: seems a little weird to be calling the class within itself. Why not just use self?

colman-hartinger:
Just ran an optimization and got a table back.
 
1. I think it could be useful to display the different infill densities of the printable and non printable meshes
2. It would be nice if the bar to drag the window around was a bit bigger and maybe a different color, it took me a while to find it and I knew it was there. 

But besides those two things and the previous comments I made in the code everything seems to work how you expect and this is a pretty cool feature

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: start

bradymadams:
I ran the code this morning. The table looks great.

1. Agree with Colman on the bar that you need to click to drag it. It's not obvious where it is.
2. When moving between stages it's great that the table is visible when coming back to the Smart Slice stage. Is it possible, though, to restore the position and size?
3. It's pretty slow moving between rows. If we can't speed it up I think a "Loading..." message would be nice so I know it's working and I don't try to keep clicking the same row.

Probably beyond the scope of this ticket, but do we have plans to maybe implement an informational popup for each row so I could hover and see what the print settings are for each row? That would be cool to be able to quickly move through them and see how they're different without selecting it and inspecting all of the print settings.

dougkenik:
> I ran the code this morning. The table looks great.
> 
> 1. Agree with Colman on the click and drag bar - it's not obvious where it is.
> 2. When moving between stages it's great that the table is visible when coming back to the Smart Slice stage. Is it possible, though, to restore the position and size?
> 3. It's pretty slow moving between rows. If we can't speed it up I think a "Loading..." message would be nice so I know it's working and I don't try to keep clicking the same row.
> 
> Probably beyond the scope of this ticket, but do we have plans to maybe implement an informational popup for each row so I could hover and see what the print settings are for each row? That would be cool to be able to quickly move through them and see how they're different without selecting it and inspecting all of the print settings.

@bradymadams & @colman-hartinger 
1. The cursor becomes a "move" cursor when it enters a transparent rectangle. I'm not 100% sold on making this rectangle a different color, so I'm interested in feedback. I can also make the rectangle bigger, if that helps?
2. Restoring the position and size would require us to save the position in the Cura Preferences (that's how they do this for other dialogs). We can do it, but it's going to take an unknown amount of work, so I would suggest putting it past MVP?
3. I'll see what I can do - it hangs when it updates the scene with the print properties and (mostly) mod meshes. We might be able to insert the mod meshes in a different thread?

Agreed on the hover tooltip - the settings update automatically in the expandable menu, but it would be good to have tooltips to see differences. Can you put an enhancement in the backlog?

bradymadams:
> @bradymadams & @colman-hartinger
> 
> 1. The cursor becomes a "move" cursor when it enters a transparent rectangle. I'm not 100% sold on making this rectangle a different color, so I'm interested in feedback. I can also make the rectangle bigger, if that helps?
> 2. Restoring the position and size would require us to save the position in the Cura Preferences (that's how they do this for other dialogs). We can do it, but it's going to take an unknown amount of work, so I would suggest putting it past MVP?
> 3. I'll see what I can do - it hangs when it updates the scene with the print properties and (mostly) mod meshes. We might be able to insert the mod meshes in a different thread?
> 
> Agreed on the hover tooltip - the settings update automatically in the expandable menu, but it would be good to have tooltips to see differences. Can you put an enhancement in the backlog?

1. A transparent bar doesn't feel standard to me. Is it not possible to drag by the colored bar that is already there?
2. Cura preferences would be if we wanted to save it between application states. We could easily save it in memory somewhere else. So the position and size would maintain within the current process, but restarting Cura would lose it. That seems reasonable to me.
3. I feel like trying to thread that will cause way more headache than just adding a message about loading.

Yes, I'll add an issue.

dougkenik:
> > @bradymadams & @colman-hartinger
> > 
> > 1. The cursor becomes a "move" cursor when it enters a transparent rectangle. I'm not 100% sold on making this rectangle a different color, so I'm interested in feedback. I can also make the rectangle bigger, if that helps?
> > 2. Restoring the position and size would require us to save the position in the Cura Preferences (that's how they do this for other dialogs). We can do it, but it's going to take an unknown amount of work, so I would suggest putting it past MVP?
> > 3. I'll see what I can do - it hangs when it updates the scene with the print properties and (mostly) mod meshes. We might be able to insert the mod meshes in a different thread?
> > 
> > Agreed on the hover tooltip - the settings update automatically in the expandable menu, but it would be good to have tooltips to see differences. Can you put an enhancement in the backlog?
> 
> 1. A transparent bar doesn't feel standard to me. Is it not possible to drag by the colored bar that is already there?
> 2. Cura preferences would be if we wanted to save it between application states. We could easily save it in memory somewhere else. So the position and size would maintain within the current process, but restarting Cura would lose it. That seems reasonable to me.
> 3. I feel like trying to thread that will cause way more headache than just adding a message about loading.
> 
> Yes, I'll add an issue.

1. We cannot use the colored bar on top because those items are reserved for clicks to "sort" the column. We would have to change the color of the rectangle on top if we wanted it to be more prominent. I'm still not 100% sold on this though, because I like the current style
2. I can look into adding it to the proxy or to the ResultsTable. That way we can save the location during an instance of Cura no problem
3. Let me see if I can change the cursor to a rotating disk while we load

SmartSlicePlugin/SmartSliceCloudStatus.py
dougkenik: we want it to be a static method which is callable. I don't believe you can use `self`

dougkenik:
I think all requests besides the header color are done. I'd like more feedback on that

SmartSlicePlugin/stage/ui/ResultTable.py
```diff
@@ -0,0 +1,333 @@
...
+
+        self.updateDisplaySignal = Signal() # Tells the owner of the table when to  update the display (like when a row is clicked)
+        self.resultsUpdated = Signal()
+
+        self._xLocation = 0
+        self._yLocation = 0
+        self._locationSet = False
+
+        self._height = 0
+        self._heightSet = False
```
bradymadams, line 69: I think it would be more "pythonic" to use `None` on the location and height attributes to indicate they're not set. Not sure if this makes it more difficult use in the QML though..?

```python
self._xLocation = None # Optional[int]
self._yLocation = None # Optional[int]
self._height = None # Optional[int]
```

dougkenik: I can look into it, I'm not sure if the QML would like it or now. In any event, I modified how this is done in #172 and can try it there

Issue #131, opened 2020-05-15
Consolidate popup QML formatting into a function


dougkenik:
We have a lot of duplicated code for formatting the popup text. Let's consolidate into a function

PR #171, opened 2020-07-24
R20.1 cancel job

colman-hartinger:
All the functionality is there, the styling using a UM message seems a little off though


colman-hartinger:
msg

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -359,12 +362,54 @@ def _createTokenFile(self):
         with open(self._token_file_path, "w") as token_file:
             json.dump(self._token, token_file)
 
+    def cancelJob(self):
+        self._client.smartslice_job_abort(self._job_id)
+        self._queue_msg = None
+
+    def my_job_status(self, job: pywim.http.thor.JobInfo):
```
jake-kirkland-teton, line 369: So we are attempting to keep consistent with the code style that Cura uses, and they do camelCased methods and underscored all_lower_case variable names. So this should be myJobStatus to keep things the same. Been working on doing this one myself.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -359,12 +362,54 @@ def _createTokenFile(self):
...
+
+    def queueMsg(self):
+        if not self._queue_msg:
+            self._queue_msg = Message(
+                text="You already have a job running from a different "
+                     "instance \n of Smart Slice. This job will be put in "
+                     "the queue and \n started once the previously "
+                     "submitted jobs have completed",
+                lifetime=600,
+                dismissable=False
```
jake-kirkland-teton, line 381: On this one, since the message is rather important, and it could possibly be in a situation where the user starts a job, and then walks away for longer than 10 minutes, assuming it might take longer then that and they come back to a failed job and no explanation; should we make this lifetime=0 and dismissable=true?

colman-hartinger: Going with the mockup we want dismissable to be False. The popup should only take a few seconds to show after hitting validate if another job is running.

colman-hartinger:
@dougkenik 

1. resolved
2. resolved
3. was already in place but fixed for edge cases
4. It could be worth creating our own QML for pop up messages. This could possibly be added to ticket #131 
5. Can you elaborate on how to duplicate these results I haven't seen this happening. With the previous code optimization jobs weren't getting canceled correctly on the server, so that could've been the problem. Now thats fixed along with a couple edge cases

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -347,12 +350,30 @@ def _createTokenFile(self):
         with open(self._token_file_path, "w") as token_file:
             json.dump(self._token, token_file)
 
+    def cancelJob(self):
+        if self._job_id:
+            self._client.smartslice_job_abort(self._job_id)
+            self._job_id = None
```
bradymadams, line 356: Tracking the job id in this class is problematic. It introduces state that needs to be managed between two different classes. I'd recommend removing the job id as an attribute in this class and sending the id as an argument to this function, instead.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -368,8 +389,10 @@ def submitSmartSliceJob(self, cloud_job, threemf_data):
                                                         pywim.http.thor.JobInfo.Status.aborted,
                                                         pywim.http.thor.JobInfo.Status.finished
                                                         ):
+
             self.job_status = task.status
-            thor_status_code, task = self._client.smartslice_job_wait(task.id)
+            thor_status_code, task = self._client.smartslice_job_wait(task.id, callback=self.myJobStatus)
```
bradymadams, line 394: Create a new object that can be called and send that as the callback. Using a function in this class means this class has to manage more state. It also would not be possible to use this class for more than one job if we ever needed to in the future.

```python
class JobStatusTracker:
    def __init__(self, ..):
        # store whatever you need here

   def __call__(self, job : JobInfo) -> bool:
        #

tracker = JobStatusTracker()

# tracker can be sent to smartslice_job_wait

File: SmartSlicePlugin/SmartSliceCloudConnector.py (hunk 3 above)
Comment (colman-hartinger) [reply SmartSlicePlugin/SmartSliceCloudConnector.py:356]: Do you have any recommendations on how to pass the unique job id to `SmartSliceCloudConnector`? My initial inclination is to send it via a signal but that also seems kind of messy

Comment (jake-kirkland-teton) [reply SmartSlicePlugin/SmartSliceCloudConnector.py:356]: Would it be smart to attach it to the SmartSliceJob class instead?

Comment (bradymadams) [reply SmartSlicePlugin/SmartSliceCloudConnector.py:356]: I'm not sure if we're doing this yet or not, but I would store the job id in the `SmartSliceCloudJob` object that gets added to the `SmartSliceCloudConnector._jobs`. Then you'll have access to it in the `SmartSliceCloudConnector` class where `cancelJob` gets called.

Comment (jake-kirkland-teton) [reply SmartSlicePlugin/SmartSliceCloudConnector.py:356]: We are not. There is a job id variable currently used by this class that I believe is still required, so we could add the new one, call it "api_job_id" or something.

Comment (bradymadams) [reply SmartSlicePlugin/SmartSliceCloudConnector.py:356]: Okay, that's fine by me. As a side note I think that _jobs dictionary design needs to be cleaned up a little, but that can be done at a later time.

The important piece is to store the info along with the job (SmartSliceCloudJob) and not in parallel in another class. So I think you guys are on the right track.

**Comment by dougkenik** [comment]
> @dougkenik
> 
> 1. resolved
> 2. resolved
> 3. was already in place but fixed for edge cases
> 4. It could be worth creating our own QML for pop up messages. This could possibly be added to ticket #131
> 5. Can you elaborate on how to duplicate these results I haven't seen this happening. With the previous code optimization jobs weren't getting canceled correctly on the server, so that could've been the problem. Now thats fixed along with a couple edge cases

4. Let's add an enhancement to the backlog to migrate our error messages to a reusable Teton message component
5. I think this will likely be fixed now that you are sending a cancellation signal

**Comment by dougkenik** [comment]
Ok. This looks so much better. Some requests:

1. The max queue message is missing a "your" - "You have exceeded the maximum allowable number of queued jobs. Please cancel a queued job or wait for `your` queue to clear"
2. Let's change the action on the max queue message to "Ok" instead of "Continue"

File: SmartSlicePlugin/SmartSliceCloudConnector.py [hunk 5]
```diff
@@ -207,6 +216,22 @@ def __init__(self, connector) -> None:
 
         self.job_type = pywim.smartslice.job.JobType.optimization
 
+
+class JobStatusTracker:
+    def __init__(self, connector, status) -> None:
+        self._previous_status = status
+        self.connector = connector
+
+    def __call__(self, job: pywim.http.thor.JobInfo) -> None:
```
bradymadams, line 225: This function needs to return a `bool` which indicates if the job should be aborted or not.

For reference, this is where the callback gets called: https://github.com/tetonsim/pywim/blob/r20.1/pywim/http/thor.py#L366-L369

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -67,6 +68,14 @@ def __init__(self, connector) -> None:
         #TODO: Get API connection
         self._client = self.connector.api_connection
 
+    @property
+    def api_job_id(self):
+        return self._api_job_id
+
+    @api_job_id.setter
+    def api_job_id(self, job_id):
+        self._api_job_id = job_id
```
bradymadams, line 77: If the `@property` is just a direct proxy to the attribute I would recommend just removing the property and renaming the attribute to omit the leading underscore.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -347,12 +372,20 @@ def _createTokenFile(self):
...
     # If the user is correctly logged in, and has a valid token, we can use the 3mf data from
     #    the plugin to submit a job to the API, and the results will be handled when they are returned.
     def submitSmartSliceJob(self, cloud_job, threemf_data):
         thor_status_code, task = self._client.new_smartslice_job(threemf_data)
 
-        Logger.log("d", "API Status after post'ing: {}".format(thor_status_code))
+        job_status_tracker = JobStatusTracker(self.connector, self.connector.status)
+
+        if hasattr(task, "id"):
+            cloud_job.api_job_id = task.id
```
bradymadams, line 386: Can you just set this below where you know `task` will have the `id` attribute? Around line 404.

SmartSlicePlugin/SmartSliceCloudConnector.py
bradymadams: To expand on the above - you can just always return `False` here if you don't want to use this mechanism for cancelling the job.

bradymadams:
Will merge pending an approval from either @dougkenik or @jake-kirkland-teton

PR #172, opened 2020-07-25
R20.1 bc style

dougkenik:
Ok.... this has quite a few changes which I'll walk through:

1.  I fixed the styling and behavior of the multiple loads / bcs dialog to match spec, and fix some issues found during testing
2. Created property tracking for the multiple loads and bcs
2. I also fixed the styling of the load dialog to match spec, as well as made it draggable in the screen
3. I created a base component class for dialogs which stores information in-memory (position, height) and assigned that to both the load dialog and the resultstable dialog. I think this will serve well for future enhancements
4. I fixed some minor issues found in the round of testing for wording and some other functionality

@colman-hartinger - please take a first crack at it


dougkenik:
Note that I merged the current Result Table branch into this. We'll need to merge that before this one, so I can merge r20.1 into this

dougkenik:
> Overall the styling looks way better. There were a couple bugs I noticed.
> 
> 1. The table where you enter the load magnitude and direction will move up when a load is added until its touching the top of the screen. I personally think that being anchored to the top of the loads and bcs popup is the best place for it
> 2. If you delete the first item in the list and then add another load or bc the incrementing is off, resulting in the possibilty of having 2 loads or bcs of the same name
> 3. The save functionality doesn't seem to be working as expected anymore. When you save a project with multiple loads and bcs and then load it back in only the first load and bc are loaded

Good review:

1. I think I fixed this - great catch
2. Another good catch - should be fixed
3. I cannot reproduce. Can you explain the steps to reproduce?

colman-hartinger:
never mind on 3 I tried to reproduce this morning and everything was working as expected

dougkenik:
> Overall I think the interface is very clean and usable. I like it.
> There are a couple minor inconveniences I noticed that are annoying at worst that would be nice if they are easy fixes, but not critical to be changed:
> 
> 1. The scroll for adding more anchors/forces once you fill the popup is really nice and clean, but is there a way to make it auto scroll when a new one is added to the bottom?
> 2. Can we add an onClicked() to the popup rectangle? If I recall correctly, it doesn't have to do anything, but without it, it currently doesn't catch all the mouse clicks, so if you click between the buttons, or between the list items for anchors/forces, the popup dismisses and you have to click back on the model, then click a force/anchor again.

I think I can do number 1. I'm lost on number 2 though - can you elaborate?

jake-kirkland-teton:
image
This is not the most amazing image, but if you click on where the red lines are, the popup will disappear, so if say, you go to click on Anchor 1, but you aim a little low or high, the popup dismisses.

jake-kirkland-teton:
Actually since it is in an Item, you would have to add a MouseArea just inside the Item, but that does mostly take care of the issue. I can commit it if you wanted to take a look.

dougkenik:
> Actually since it is in an Item, you would have to add a MouseArea just inside the Item, but that does mostly take care of the issue. I can commit it if you wanted to take a look.

What does "mostly" mean?

jake-kirkland-teton:
It just doesn't quite extend the MouseArea to the pop-up edges. The margins aren't covered.

dougkenik:
I think that's fine

jake-kirkland-teton:
Yeah. It's much better with the MouseArea, and the small amount of margins is no big deal.

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -148,7 +151,10 @@ def deactivate(self):
 
     @pyqtSlot()
     def add(self):
-        N = len(self._bcs)+1
+        if (len(self._bcs) == 0):
+            N = 1
+        else:
+            N = (int)(self._bcs[-1].getName().split(" ")[1]) + 1
```
bradymadams, line 157: Remove parens around `int`.

SmartSlicePlugin/select_tool/SmartSliceSelectTool.qml
```diff
@@ -90,106 +93,203 @@ Item {
...
+                onPressed: {
+                    clickPos = Qt.point(mouse.x, mouse.y);
+                    dragging = true
+                }
+                onPositionChanged: {
+                    if(dragging)
+                    {
+                        var delta = Qt.point(mouse.x - clickPos.x, mouse.y - clickPos.y);
+                        if (delta.x !== 0 || delta.y !== 0)
+                        {
```
bradymadams, line 176: Open curly brace on previous line.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -56,6 +57,8 @@ def initialize(self, parent: SceneNode):
 
             self._interactive_mesh = makeInteractiveMesh(mesh_data)
 
+        self.rootChanged.emit()
```
bradymadams, line 60: Emitting this signal here could be problematic. It's not definitive that anytime a Root is created that it becomes the new Root in the Scene that we want to track. I would emit it where the Root is created and actually added to the relevant SceneNode.

dougkenik: ok. Where does that happen? I tried to tap in to a signal the Cura Scene emits, but nothing ever emitted. Do we do this somewhere?

bradymadams: Never mind. The way the git diff showed up I thought you were emitting the signal in `__init__`. You have it exactly where it should be. My only request would be to send the `parent` as an argument in the signal emission. You can then use the `parent` directly in the connected functions, rather than searching for the `Root`.

dougkenik:
Ok, I added a mouse area to reduce the headache of the Tool dismissing on random clicks. I wasn't able to figure out how to set the scroll position to the end when a new object is added, so we can create an enhancement (I think I know how to do it, but there is a decent amount of work to restyle the QML)

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -59,7 +60,7 @@ def __init__(self, connector):
         self._selected_material = SmartSliceProperty.SelectedMaterial()
         self._scene = SmartSliceProperty.Scene()
         self._root = SmartSliceProperty.SmartSliceSceneRoot()
-        self._onRootChanged()
+        self._onRootChanged(None)
```
bradymadams, line 63: This call doesn't do anything, right?

dougkenik: No, I could theoretically remove it if you want

bradymadams: Let's remove it then

dougkenik:
I found a bug I'm trying to figure out - please hold

dougkenik:
> Everything looks pretty good to me.
> 
> One last thing that could be an annoyance for the user, is there a way we can change the checks on Factor of Safety to on editing finished or on accept? Currently you can't clear the box, so if you want to change the value from say, 1 to 5, you have to type 15, and then erase the 1.

@jake-kirkland-teton - I put this in place because a user could enter something below 1.0. For some reason the validator would still allow values under 1.0, even though the bottom was set to 1.0. Any thoughts?

jake-kirkland-teton:
Ah. I was kinda afraid that was the snag. Hmm. I will have to play with it a little and see if I can find a good solution.

PR #178, opened 2020-07-28
dynamic server loading

colman-hartinger:


SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -537,10 +553,12 @@ def __init__(self, proxy: SmartSliceCloudProxy, extension):
 
         self.extension = extension
 
+        self.plugin_data = PluginMetaData()
```
dougkenik, line 556: Maybe we should have defaults in the code, and overwrite if values are present?

SmartSlicePlugin/plugin.json
```diff
@@ -5,5 +5,7 @@
     "version": "20.1.14.0",
     "description": "Optimizes your model and slicing properties based on physical boundary conditions set by the user.",
     "supported_sdk_versions": ["7.0.0", "6.3.0"],
-    "i18n-catalog": "smartslice"
+    "i18n-catalog": "smartslice",
+    "APIURL": "test.smartslice.xyz",
+    "APICluster": "enter value here"
```
dougkenik, line 10: What happens if API cluster is left as-is in this file? Are we going to pick it up and set it to this value, or will it be None?

bradymadams: APICluster should default to `None`. So I'd prefer that we omit it, by default.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -537,10 +553,12 @@ def __init__(self, proxy: SmartSliceCloudProxy, extension):
...
+
         # Application stuff
         self.app_preferences = Application.getInstance().getPreferences()
         self.app_preferences.addPreference(self.http_protocol_preference, "https")
-        self.app_preferences.addPreference(self.http_hostname_preference, "test.smartslice.xyz")
+        self.app_preferences.addPreference(self.http_hostname_preference, self.plugin_data.api_url)
         self.app_preferences.addPreference(self.http_port_preference, 443)
```
bradymadams, line 562: We should probably consider removing these preferences? Do you have a feel for what takes precedence? I'd rather there just be a single source of truth for this info. Storing it in more than one place is going to frustrate someone who wants to change this and doesn't realize their change isn't being respected because there is another configuration that is taking precedence.

SmartSlicePlugin/plugin.json
bradymadams, line 10: Let's format the info a little differently so it's not flat and we can expand on it, if necessary:

```json
"smartSliceApi": {
    "url": "https://test.smartslice.xyz",
}
```

bradymadams: Also please include the protocol on the URL: `https://`

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: isn't the default supplied value already in plugin.json and we're just reading in that value wherever need be?

dougkenik: I guess the question would be, what if it isn't supplied for some weird reason, like a user mucks with something and screws it up?

colman-hartinger:
@bradymadams should the protocol be included? I thought that  was specified in `        self.app_preferences.addPreference(self.http_protocol_preference, "https")
`

bradymadams:
> @bradymadams should the protocol be included? I thought that was specified in ` self.app_preferences.addPreference(self.http_protocol_preference, "https")`

The point of the URL configuration is to specify the entire URL, including the protocol. For example:

`https://test.smartslice.xyz:443`

We can then parse out the pieces. We should only store this in one place, so I think we should remove any check against the cura preferences file.

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: If that's something to worry about then I dont think any default value would be safe. But it should be painless to add some dummy variables to the PluginMetaData class

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: Ok I get what to do now. I'll get rid of all the preferences and just load everything from the class

PR #180, opened 2020-07-28
Bugfix

jake-kirkland-teton:
This fixes the bug where you can login with a different user and then get errors about wrong user/password even when you are using the correct login credentials for a different account.


jake-kirkland-teton:
Using the package installer it is possible to have a valid token for a different account, and still be required to login. And if you attempt to login with a different account, it will fail, even if your email/password is correct. This fix just makes sure the token gets cleared if it's checked and ends up not matching.

PR #187, opened 2020-08-04
R20.1 progress bar

### LINKED ISSUE #174: Progress Bar
Provide a progress bar during validation and optimization bu using the progress updater on our AWS system. Design is here:
https://tetoncompositescom.sharepoint.com/:p:/s/TetonSimulation/EWUJCII_PINNv142Gs6kk94B_KK0ctJsuZY8Wfqu0xaLPw?e=2m3nI3

jake-kirkland-teton:
Progress bar for showing progress on a validation or an optimize. Currently validations happen to go from 0 to 100, so the bar never fills, but it is fully functioning for Optimizations.

Closes #174


SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -57,6 +57,9 @@ def __init__(self) -> None:
         self._sliceInfoOpen = False
         self._errors = {}
 
+        self._jobProgress = 0
```
dougkenik, line 60: We haven't really been following Cura standards on naming things within the plugin, and we need to start.

Any variable contained within a class should be named with underscores. e.g.` self._job_progress`. If the variable is internal to that class, then a leading underscore should be used, otherwise not. All function names should use Camel Case e.g. `progressBarVisibleChanged`. Let's comb through your code and make sure we're following these standards with the variables you added.

The naming convention is obviously different for QML

SmartSlicePlugin/stage/ui/SmartSliceMain.qml
```diff
@@ -16,12 +16,11 @@
 //  API Imports
 import QtQuick 2.7
 import QtQuick.Controls 2.2
-import QtQuick.Controls 1.4 as QQC1
```
dougkenik, line 19: I believe if you import 1.4 before 2.2, you won't need a different name, but I could be mistaken.

SmartSlicePlugin/stage/ui/SmartSliceMain.qml
```diff
@@ -106,6 +105,21 @@ Item {
                         text: SmartSlice.Cloud.sliceStatus
                     }
 
+                    UM.ProgressBar
+                    {
```
dougkenik, line 109: Put the open curly bracket on the previous line `UM.ProgressBar {`

SmartSlicePlugin/SmartSliceCloudProxy.py
jake-kirkland-teton: Ahhh crap, I was just telling Colman about this one. And then I forgot myself. Gotcha. I'll get it fixed.

SmartSlicePlugin/stage/ui/SmartSliceMain.qml
jake-kirkland-teton: This one was a leftover from a few changes ago actually, not exactly sure how it didn't get cleared that time, I remember changing it then. Just removing it again, not currently needed.

SmartSlicePlugin/stage/ui/SmartSliceMain.qml
jake-kirkland-teton: Gotcha.

jake-kirkland-teton:
I can't answer number 1, that is a Brady question for sure. But number 2 would be incredibly easy if we can't do 1.

bradymadams:
> I can't answer number 1, that is a Brady question for sure. But number 2 would be incredibly easy if we can't do 1.

Here's all of the possible values for job status:

https://github.com/tetonsim/pywim/blob/r20.1/pywim/http/thor.py#L85-L92

As it is, this PR looks good to me, but I'm not sure if @dougkenik wants some changes based off the available status values?

dougkenik:
I was just thinking the progress indicator could take into account the time it takes to spin up the system for the job. Maybe this is a ticket for a later day though.

@jake-kirkland-teton - let's remove the progress bar from the validate step and create a ticket to revisit the progress status

bradymadams:
> I was just thinking the progress indicator could take into account the time it takes to spin up the system for the job. Maybe this is a ticket for a later day though.

Yeah, possibly. Although, there are some tickets in aws-infra to improve the auto-scaling. I think it's possible we could improve auto-scaling to where the "spin-up" time isn't as long as it currently is.

dougkenik:
> > I was just thinking the progress indicator could take into account the time it takes to spin up the system for the job. Maybe this is a ticket for a later day though.
> 
> Yeah, possibly. Although, there are some tickets in aws-infra to improve the auto-scaling. I think it's possible we could improve auto-scaling to where the "spin-up" time isn't as long as it currently is.

Ok. I think the actions I outlined are appropriate for now then

jake-kirkland-teton:
Alright, Issue created and Validate bar removed. Should be all set!

PR #195, opened 2020-08-07
no optimization resultss helper

colman-hartinger:
Message looks a little weird without  a title but matches the mockups pretty welll


dougkenik:
You have a test protocol?

colman-hartinger:
Thats a little strange, I'll reformat the html so it will stay in the right format. Looked great on my computer

colman-hartinger:
I think I fixed things, changed the message to pure html and no `\n` which is what I think was causing you to get the strange formatting. Also I added the message to display if you switch to the smart slice stage without a printable node on the build plate so you can check the message without running an optimization.

colman-hartinger:
this is what I'm seeing 
noopt

dougkenik:
I'm getting this - which is better. I still want an extra line between the paragraphs for readability:

image

colman-hartinger:
alright Should be fixed now. I had the message display if you tried to switch to the smart slice screen without a printable model for troubleshooting and that's where I was displaying the title should be in both messages now though.

dougkenik:
I think it needs work?

image

colman-hartinger:
alright I think all the spaces are in the correct place now. It's strange because nothing has changed with how my computer is rendering the message

dougkenik:
image

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -862,11 +862,25 @@ def _onJobFinished(self, job):
         else:
             if self.status != SmartSliceCloudStatus.ReadyToVerify and self.status != SmartSliceCloudStatus.Errors:
                 self.prepareOptimization() # Double Check Requirements
+                results = self.extension.getNoOptimizationResults()["structural"]
```
dougkenik, line 887: Is there ever a time when no results come back?

colman-hartinger: Maybe if the backend couldn't evaluate the solid model for some reason. I can surround the logic in an if statement so it won't break given an unexpected error

dougkenik: > Maybe if the backend couldn't evaluate the solid model for some reason. I can surround the logic in an if statement so it won't break given an unexpected error

I just want to make sure we aren't going to issue a breaking error if it doesn't happen

colman-hartinger: better safe than sorry, just added an error check

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -884,11 +884,34 @@ def _onJobFinished(self, job):
         else:
             if self.status != SmartSliceCloudStatus.ReadyToVerify and self.status != SmartSliceCloudStatus.Errors:
                 self.prepareOptimization() # Double Check Requirements
-                Message(
-                    title='SmartSlice',
-                    text=i18n_catalog.i18nc("@info:status", "SmartSlice was unable to find a solution")
-                ).show()
-
+                if self.extension.getNoOptimizationResults():
```
bradymadams, line 887: Not sure I understand this. This function returns the first item from a list - won't this always be true then?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -884,11 +884,34 @@ def _onJobFinished(self, job):
...
+                             "<p>Solid Print:</p> <p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; "
+                             "Maximum Displacement:"
+                             "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; "
+                             "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; %.2f</p>"
+                             "<p></p>"
+                             "<p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; "
+                             "Minimum Factor of safety:"
+                             "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; "
+                             "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; %.2f</p>" %
+                             (results["max_displacement"], results["min_safety_factor"]),
```
bradymadams, line 905: All the `&nbsp`s make this hard to read and will be difficult to maintain. I'm not real familiar with how Qt renders HTML, but if it respects `style` attributes then we should use CSS instead to format this block of text. Do you know if style attributes get applied?

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -185,6 +185,10 @@ def _saveState(self, writing_workspace=False):
             self._storage.setEntryToStore(plugin_id=self.metadata.id, key='results', data=None)
             self._storage.setEntryToStore(plugin_id=self.metadata.id, key='selectedResult', data=None)
 
+    def getNoOptimizationResults(self):
+        all_data = self._storage.getPluginMetadata(self.metadata.id)
+        return all_data["results"]["analyses"][0]
```
bradymadams, line 190: This is getting the result that is stored in the plug-in metadata. Is that what we want to get right after the analysis has finished?

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: got rid of it. thought it would be safer to put some kind of error check but didn't think about that

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: sounds good, I'll use css formating

SmartSlicePlugin/SmartSliceExtension.py
colman-hartinger: Ya you are right this is not the right place this is just getting the old validation data not the solid model evaluation. I need to figure out where the solid model results get sent before the are cleared

SmartSlicePlugin/SmartSliceCloudConnector.py
colman-hartinger: so Qt did  a bad job a rendering css styling but the `pre` tag seemed to work just as well

colman-hartinger:
@dougkenik The font should be fixed and the status is changed to prepare optimization now so it doesn't stay in the optimizing tab. Should I switch the status to prepare validation since in the prepare optimization it will display the validation results?

@bradymadams how Qt renders html and css is kind of strange a small percent of the styling that was rendered in my browser would be rendered by Qt, even other html whitespace operators wouldn't be rendered. I think this is the cleanest solution without making our own messaging system

dougkenik:
I would switch it back to "Validate," since the system cannot find a viable solution

bradymadams:
> @bradymadams how Qt renders html and css is kind of strange a small percent of the styling that was rendered in my browser would be rendered by Qt, even other html whitespace operators wouldn't be rendered. I think this is the cleanest solution without making our own messaging system

Can you put the number in a nested `<p>` and use another `text-indent` on it so you don't need the nbsps?

colman-hartinger:
> > @bradymadams how Qt renders html and css is kind of strange a small percent of the styling that was rendered in my browser would be rendered by Qt, even other html whitespace operators wouldn't be rendered. I think this is the cleanest solution without making our own messaging system
> 
> Can you put the number in a nested `<p>` and use another `text-indent` on it so you don't need the nbsps?

No a nested <p> would have the number on a new line and text-indent also only works on new lines. I tried a <span> with a word spacing style which seemed like the best option but Qt wouldn't render that

bradymadams:
> No a nested `<p>` would have the number on a new line and text-indent also only works on new lines. I tried a `<span>` with a word spacing style which seemed like the best option but Qt wouldn't render that

Okay. I suppose this will work for now then. It feels like using HTML as text in a simple message box probably isn't the best long term approach then. In the future, if we come back to adjusting this box we should probably look into moving it into QML.

colman-hartinger:
I was thinking the same thing, this works, but our own messaging system would be cleaner

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -11,7 +11,7 @@
 
 import pywim  # @UnresolvedImport
 
-from PyQt5.QtCore import pyqtSignal, pyqtProperty, pyqtSlot
+from PyQt5.QtCore import pyqtSignal, pyqtProperty, pyqtSlot, QCoreApplication
```
bradymadams, line 14: QCoreApplication unused

bradymadams:
Will merge pending approval from @dougkenik

dougkenik:
Minor formatting requests:

1. Period at the end of the first paragraph
2. Underline "Solid Print" 
3. Put a new line after Solid Print and before Minimum Safety Factor
4. Italicize "Minimum Safety Factor" and "Maximum Displacement" and also italicize the values

image

PR #206, opened 2020-08-11
R20.1 load handler

dougkenik:
Sorry about this one - lots of stuff going on. I wasn't sure about where to put the tool handles, and ended up deciding putting them with the HighlightFace was the best option, even though this goes against the typical Cura standard - the reason being, that Tools typically are associated with a body, and we are associating this with a face.

So I migrated the arrow to a Tool Handle, and created the rotator to also be a tool handle. I tried to make the arrow a bit "dynamic" where i did not need 2 for the different directions (push/pull), but I couldn't figure out how Cura recognizes dynamic SceneNodes - the positions are set in very weird positions depending on when you add them to the scene.

Also, I had to hack something together to get the selection working like we need for the UX. This forced me to turn on/off SelectFace mode in a lot of places to ensure we were capturing user intent (otherwise there is a nasty bug in Cura). 

To test, you'll have to change the `surface_type` and `direction_type` properties until we get the buttons implemented. Property tracking and saving/opening should work

As an FYI - there is another branch taken off of this, so we should identify the best merging strategy


SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -260,156 +232,320 @@ def pywimBoundaryCondition(self, step: pywim.chop.model.Step, mesh_rotation: Mat
...
+        self.activeArrow.rotate(matrix, SceneNode.TransformSpace.World)
+        self.inactiveArrow.rotate(matrix, SceneNode.TransformSpace.World)
+
+        self.activeArrow.setPosition(self._rotator.center)
+        self.inactiveArrow.setPosition(self._rotator.center)
+
+        self.inactiveArrow.setEnabled(False)
+
+        self.activeArrow.direction = matrix.rotate(self.activeArrow.direction)
+        self.inactiveArrow.direction = matrix.rotate(self.inactiveArrow.direction)
```
bradymadams, line 336: why is inactiveArrow enabled and then disabled?

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -260,156 +232,320 @@ def pywimBoundaryCondition(self, step: pywim.chop.model.Step, mesh_rotation: Mat
...
+        if not face.triangles:
+            return HighlightFace.SurfaceType.Unknown
+        elif len(self._interactive_mesh.select_planar_face(face.triangles[0]).triangles) == len(face.triangles):
+            return HighlightFace.SurfaceType.Flat
+        elif len(self._interactive_mesh.select_concave_face(face.triangles[0]).triangles) == len(face.triangles):
+            return HighlightFace.SurfaceType.Concave
+        elif len(self._interactive_mesh.select_convex_face(face.triangles[0]).triangles) == len(face.triangles):
+            return HighlightFace.SurfaceType.Convex
+
+        return HighlightFace.SurfaceType.Unknown
```
bradymadams, line 551: Empty line at EOF.
Will the rest of the code handle an `Unknown` face type okay? We could offload the face type to some class in pywim.chop so it gets saved, but that kind of adds a restriction of face types in chop. Maybe something to think about.

colman-hartinger:
when I  tried to run the code and got this error on opening smart slice tab

```
SmartSliceSelectTool.py", line 170, in _getSelectedTriangles
    axis = smart_slice_node._interactive_mesh.planar_axis(selected_triangles)
AttributeError: 'Mesh' object has no attribute 'planar_axis'

```

dougkenik:
> when I tried to run the code I got this error on opening smart slice tab and cura crashed
> 
> ```
> SmartSliceSelectTool.py", line 170, in _getSelectedTriangles
>     axis = smart_slice_node._interactive_mesh.planar_axis(selected_triangles)
> AttributeError: 'Mesh' object has no attribute 'planar_axis'
> ```

Let's hold off on the review for now. I have to change some things in pywim and update the code base

SmartSlicePlugin/select_tool/RotateToolHandle.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2015 Ultimaker B.V.
+# Uranium is released under the terms of the LGPLv3 or higher.
```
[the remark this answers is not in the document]
dougkenik: It was not copied. - Just leftover from experimenting

SmartSlicePlugin/stage/SmartSliceScene.py
dougkenik: You have to activate it in order to rotate it. I deactivate it again so it is invisible

SmartSlicePlugin/stage/SmartSliceScene.py
dougkenik: I don't think it will ever happen, so it shouldn't be an issue. I put it in for future use if needed

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -255,35 +255,67 @@ def changed(self) -> bool:
...
+            self.face.force.magnitude = self._properties.magnitude
+            self.face.force.pull = self._properties.pull
+            self.face.force.direction_type = self._properties.direction_type
+
+        self.face.surface_type = self._properties.surface_type
+        self.face.setMeshDataFromPywimTriangles(self._properties.triangles, self._properties.axis)
+
+        # Rotate the load arrow back to match
+        if isinstance(self.face, LoadFace):
+            self.face.setArrow(self._properties.direction)
```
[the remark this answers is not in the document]
dougkenik: it could be, but I think there would be a circular dependency then, because we would have to pass the `Properties` class back to the `HighlightFace` - we could move the `Properties` class to `HighlightFace`, but that seems like a very weird place to put it. Thoughts?

dougkenik:
Ok guys, we're ready to try this again. Make sure to use the `r20.1-load_type` pywim branch

SmartSlicePlugin/select_tool/LoadArrow.py
```diff
@@ -0,0 +1,133 @@
...
+
+from .LoadToolHandle import LoadToolHandle
+
+class LoadArrow(LoadToolHandle):
+    """Provides the arrow for the load direction"""
+
+    def __init__(self, parent = None):
+        super().__init__(parent)
+
+        self._name = "LoadArrow"
```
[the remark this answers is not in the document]
dougkenik: It's used in ToolHandle, which LoadArrow is eventually derived from

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -214,43 +158,74 @@ def pywimBoundaryCondition(self, step: pywim.chop.model.Step, mesh_rotation: Mat
...
+        self._axis = None
+
+        self._arrows = {
+            True: LoadArrow(self),
+            False: LoadArrow(self)
+        }
+        self._rotator = LoadRotator(self)
+        self._rotator.buildMesh()
+        for key, arrow in self._arrows.items():
+            arrow.buildMesh(pull = key)
```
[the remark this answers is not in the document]
dougkenik: No - this is iterating through the dictionary `self._arrows` - which has one arrow for a True (Pull) direction, and one arrow for a Push (False) direction. Basically we initialize use two different classes of `LoadArrow`, pointed in different directions, then build them. This dictionary is used to toggle the arrow we want, as the user selects the flip button in the UI

dougkenik:
> When running the code it seems like the way that anchor and load faces select are now very different from eachother. The anchor faces were able to pick up on the face selected instantly but the load faces sometimes would noy go to the face you select no matter how many times clicked (this is for surface_type = convex ). Another bug I found was that when surface_type = flat and load_direction = parallel the arrow would disappear after a few clicks
> 
> My initial inclination is that these bugs are caused by the event function in SmartSliceSelectTool but I could be wrong

@colman-hartinger - A few comments / questions:

1. I would expect slower behavior when selecting Concave and Convex faces - When exactly was it slow? The `event ` is only executed if the rotation handler is visible
2. You were only seeing this odd behavior with Convex faces?
3. I'll look into the load arrow problem

colman-hartinger:
>     1. I would expect slower behavior when selecting Concave and Convex faces - When exactly was it slow? The `event ` is only executed if the rotation handler is visible
> 
>     2. You were only seeing this odd behavior with Convex faces?
> 

I meant to say concave. I'll slack you a video of the difference I'm seeing between the anchor and load face selection

SmartSlicePlugin/SmartSliceProperty.py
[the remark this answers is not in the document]
bradymadams: I'm not necessarily suggesting the `SmartSliceFace` class would be passed to `HighlightFace`. But at the very least I think it would make sense to inherit from `SmartSliceFace` - one for loads (and one for anchors if necessary), and add the additional functionality for loads in `SmartSliceLoadFace`, or whatever you call it.

SmartSlicePlugin/stage/SmartSliceScene.py
bradymadams: The length check worries me a little bit. If the surface selection algo is every changed slightly and someone opens a previously saved 3MF won't this have the potential to return an `Unknown`?

dougkenik:
> Rotating the arrow is a little choppy and slow on my machine. The log output dumps a lot of info while I'm rotating - I'm assuming it's coming from `active_node.facePropertyChanged.emit()`. If I comment that out it works better for me. Can we hold off on emitting that signal until the mouse release event?
> 
> Do we know why the arrow and donut break up the part outline rendering? It doesn't seem critical, but it sort of jumps out at me while I'm rotating. If it's possible to fix that I think it would look cleaner if the break didn't happen.
> 
> image

I can definitely emit the signal on the mouse release event. The break in the outline is due to the renderer - I think we would need to write our own Selection Pass to get around it

SmartSlicePlugin/SmartSliceProperty.py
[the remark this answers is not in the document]
dougkenik: I'll look into it

dougkenik:
> > Rotating the arrow is a little choppy and slow on my machine. The log output dumps a lot of info while I'm rotating - I'm assuming it's coming from `active_node.facePropertyChanged.emit()`. If I comment that out it works better for me. Can we hold off on emitting that signal until the mouse release event?
> > Do we know why the arrow and donut break up the part outline rendering? It doesn't seem critical, but it sort of jumps out at me while I'm rotating. If it's possible to fix that I think it would look cleaner if the break didn't happen.
> > image
> 
> I can definitely emit the signal on the mouse release event. The break in the outline is due to the renderer - I think we would need to write our own Selection Pass to get around it

The break in the outline is due to a "selection mesh" that the tool handles use to cause the other mesh to highlight. I tried to reduce the padding, but I think they might always be there, unless we can find a better implementation

dougkenik:
Can you guys take a look at this again? It feels solid in my hands, except for some styling stuff on the blue boundary of the node... which I am unsure of how to fix

colman-hartinger:
It seems like whenever you click in the smart slice scene where the part isn't located you get this error 
```SmartSliceSelectTool.py", line 155, in _getSelectedTriangles
    selected_face = smart_slice_node._interactive_mesh.select_concave_face(face_id)
  File "/home/colman/Documents/cura/pywim/pywim/geom/tri.py", line 544, in select_concave_face
    tri = next(t for t in self.triangles if t.id == tri)
StopIteration```

**Comment by colman-hartinger** [comment]
also the anchor face seems to default to concave surface selection. On the concave surface of the simple lever I'm not able to just select a flat surface
sl

But everything besides those two bugs seems to be working awesome and the plugin is starting to look badass. Couldn't break anything else in the rotate tool even when mashing all the buttons and irrationally changing the direction.

**Comment by dougkenik** [comment]
> also the anchor face seems to default to concave surface selection. On the "concave" surface of the simple lever I'm not able to just select a flat surface
> sl
> 
> But everything besides those two bugs seems to be working awesome and the plugin is starting to look badass. Couldn't break anything else in the rotate tool even when mashing all the buttons and irrationally changing the direction.

Anchor faces have a separate ticket - this was an oversight on my part

**Comment by dougkenik** [comment]
> It seems like whenever you click in the smart slice scene where the part isn't located you get this error and cura crashes
> 
> ```
>     selected_face = smart_slice_node._interactive_mesh.select_concave_face(face_id)
>   File "/home/colman/Documents/cura/pywim/pywim/geom/tri.py", line 544, in select_concave_face
>     tri = next(t for t in self.triangles if t.id == tri)
> StopIteration```
> ```

I need some help reproducing this. I cannot find a way to do it. Have you pulled the latest pywim branch for this (load_type) and the latest smartslice load_handler branch?

**Comment by colman-hartinger** [comment]
> > It seems like whenever you click in the smart slice scene where the part isn't located you get this error and cura crashes
> > ```
> >     selected_face = smart_slice_node._interactive_mesh.select_concave_face(face_id)
> >   File "/home/colman/Documents/cura/pywim/pywim/geom/tri.py", line 544, in select_concave_face
> >     tri = next(t for t in self.triangles if t.id == tri)
> > StopIteration```
> > ```
> 
> I need some help reproducing this. I cannot find a way to do it. Have you pulled the latest pywim branch for this (load_type) and the latest smartslice load_handler branch?

Ya I just fetched and pulled code from both branches before testing and then tried again when I got the error. I'll slack you the complete error log and a screen cast of the steps I am taking

**Comment by colman-hartinger** [comment]
So when I put break points in any of the select face functions in pywim.geom.tri when clicking on the empy space the value for `tri` is 8388607 which is the faceID from the smart slice select tool. current surface should be none for this case which is strange. Maybe we are using different versions of cura since current_surface is called through Selection.getSelectedFace

**Comment by dougkenik** [comment]
> So when I put break points in any of the select face functions in pywim.geom.tri when clicking on the empy space the value for `tri` is 8388607 which is the faceID from the smart slice select tool. current surface should be none for this case which is strange. Maybe we are using different versions of cura since current_surface is called through Selection.getSelectedFace

This is due to the `SlectionPass ` for the `Tool `class. Basically, that number you are seeing is related to a pixel, but not a face. This happens because the `Tool` class implements `SelectionPass` in a weird way. I thought I found a way around this with some other logic I put in, but obviously not. If you can show me how to reproduce, I might be able to fix it - otherwise I will need to change directions for the SelectTool

**Comment by bradymadams** [comment]
I get some strange behavior where I can't click the rotator, and then after switching between options it eventually disappears. I have to end up adding another load and going back to the original to get it to work. See screen recording:

load-rotator-behavior

**Comment by bradymadams** [comment]
Should convex surface selection be working? I can't get it to select any faces at all. I've tried a couple of STLs.

**Comment by dougkenik** [comment]
Works for me on this one. In general, if it can't find a rotation axis, it won't highlight the face:

image

**Comment by dougkenik** [comment]
> I get some strange behavior where I can't click the rotator, and then after switching between options it eventually disappears. I have to end up adding another load and going back to the original to get it to work. See screen recording:
> 
> load-rotator-behavior

I've noticed this too. It's not capturing the mouse move event to turn on the body selection. I'll keep digging

**Comment by bradymadams** [comment]
Some more unexpected behavior. Note that after I switched from Load 2 to Load 1 I didn't click anything - all I did was move my mouse and the selected face moved.

load-rotator-behavior-2

**Comment by bradymadams** [comment]
Perhaps related to Colman's problem.

[cylinder.stl.zip](https://github.com/tetonsim/is-cura-ui/files/5097068/cylinder.stl.zip)

load-rotator-behavior-3

**Comment by bradymadams** [comment]
Sometimes the arrow disappears entirely on me and I can't get it back. This isn't the best screen recording - I can't figure out the exact steps to repro so I do a lot here, but it does show the issue towards the end.

load-rotator-behavior-4

**Comment by dougkenik** [comment]
Ok. I revamped how the selection is working, and it feels stable for me now.... give it another shot when you have a chance and let me know what issues you find.

**Comment by colman-hartinger** [comment]
So the clicking on an empty area almost seems to be fixed. 99% of the time when I click on whitespace, the only time I can get cura to crash now is if you are using the load manipulator and quickly let go and click white space the pixel value is still captured and cura crashes. 

The other bug I found was that when the build plate is cleared and another part is added you arent able to switch to the smart slice tab. This seems like a quick fix

error

**Comment by dougkenik** [comment]
> So the clicking on an empty area almost seems to be fixed. 99% of the time when I click on whitespace, the only time I can get cura to crash now is if you are using the load manipulator and quickly let go and click white space the pixel value is still captured and cura crashes.
> 
> The other bug I found was that when the build plate is cleared and another part is added you arent able to switch to the smart slice tab. This seems like a quick fix
> 
> error

Ok. I think this is fixed now

**Comment by jake-kirkland-teton** [comment]
So I found a couple more bugs, hopefully these will be fairly simple to fix:

1) If you get a successful optimization, the rotation tool and it's potential arrows end up showing on the preview tab. 
image

2) If you end up attempting to make a change while Validating/Optimizing or have results, weird behavior can occur, like if you have the anchor selected and change anchors, then click cancel, both the anchor and loads will appear. Also if you have loads selected, and change one, then cancel, sometimes you get a fatal crash, sometimes the load selection tool does not reset correctly.
image
image

**Comment by dougkenik** [comment]
Ok - @jake-kirkland-teton. I believe all of these have been addresses. Give it a shot and let me know what else you find.

**Comment by jake-kirkland-teton** [comment]
Alright. Those all appear to be fixed. And the general feel is really, really quite awesome. I'm impressed.

I am also good at breaking stuff though. So I managed to get the arrow and load tool to show up on the prepare tab now too:
image

I am also getting an issue where if I have the convex/concave tool selected, and change face to something that deselects the face, I don't get the warning about invalidating a currently active job. It just clears the job. And you can get some really bizarre behavior in addition to that as well, such as cancelling it, the button going inactive and the arrow still visible, but the circle for the angle tool disappearing. And when the button is disabled, there is no reason, but the popup still appears.
image
image

**Comment by dougkenik** [comment]
Ok. I think everything has been fixed. There were some interesting bugs in the property handler, which I think I sorted through. If you guys could look at the code and the workflow again, I would appreciate it so we can get this merged as soon as possible

**Comment by bradymadams** [comment]
I'll merge once @dougkenik is ready and pending @colman-hartinger approval.

File: SmartSlicePlugin/select_tool/SmartSliceSelectTool.qml [hunk 7]
```diff
@@ -220,319 +260,506 @@ Item {
...
+                                            horizontalCenter: parent.horizontalCenter
+                                            verticalCenter: parent.verticalCenter
+                                        }
+
+                                    Repeater {
+                                        id: tickmarks
+                                        model: loadHelperData.maxValue / loadHelperData.stepSize -1
+                                        Rectangle {
+                                            function indexHelper(index) {
+                                                console.log(model)
```
colman-hartinger, line 607: this console log doesn't seem necessary

SmartSlicePlugin/requirements_tool/SmartSliceRequirements.py
```diff
@@ -2,6 +2,7 @@
 
 from UM.Tool import Tool
 from UM.Signal import Signal
+from UM.Scene.Selection import Selection
```
colman-hartinger, line 5: Unnecessary import

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -9,14 +9,14 @@
 
 from ..utils import findChildSceneNode, getPrintableNodes
 from ..stage import SmartSliceScene
+from ..select_tool import SmartSliceSelectTool
```
colman-hartinger, line 12: unused import

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -89,29 +97,63 @@ def boundaryConditionType(self, value: int):
...
+    def surfaceType(self) -> int:
+        if self._active_node:
+            return self._active_node.surface_type.value
+        return SmartSliceScene.HighlightFace.SurfaceType.Flat.value
+
+    @surfaceType.setter
+    def surfaceType(self, value: int):
+        if self._active_node.surface_type.value != value:
+            self._active_node.surface_type = SmartSliceScene.HighlightFace.SurfaceType(value)
+            # self._active_node.facePropertyChanged.emit() # This will get emitted by connections to Selection.selectedFaceChange
```
colman-hartinger, line 148: should this comment still be here?

SmartSlicePlugin/select_tool/BoundaryConditionList.py
```diff
@@ -125,6 +167,8 @@ def data(self, index, role):
 
     @pyqtSlot(int)
     def activate(self, index=0):
+        # Selection.setFaceSelectMode(True)
+
```
colman-hartinger, line 171: comment

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -249,19 +213,163 @@ def getSurfaceType(self):
...
+            # Must be a left mouse event to select or rotate
+            if MouseEvent.LeftButton not in event.buttons:
+                return False
+
+            pixel_color = self._selection_pass.getIdAtPosition(event.x, event.y)
+
+            # We did not click the tool - we need to select the surface under it if it exists
+            # TODO - This is a little hacky.... we should implement a SelectionPass just for this Tool
+            if not pixel_color or not arrow.isAxis(pixel_color):
+                if Selection.hasSelection() and not Selection.getFaceSelectMode():
```
colman-hartinger, line 279: this seems like it can be put into a single if statement

SmartSlicePlugin/select_tool/SmartSliceSelectTool.qml
dougkenik: Good catch! I was wondering where that was coming from

SmartSlicePlugin/select_tool/BoundaryConditionList.py
dougkenik: I want to keep it there because it's a reminder not to put it back in. We were emitting a signal twice, and it was causing issues

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
dougkenik: It could, but the readability will be pretty horrendous. Thoughts?

colman-hartinger: ya I just tried it out, looks pretty gross. Probably better to keep it how it is

dougkenik:
Everything should be good here. We may want one more round of test to ensure nothing crazy is going on?

colman-hartinger:
Just pulled all the new code and tested things out. Couldn't find any bugs

PR #213, opened 2020-08-13
R20.1 about box

dougkenik:
About box styling and links for EULA, Privacy, etc.


dougkenik:
> Looks good. What is the consensus on semi-colons and QML attributes? I always use them. Looks like that's not preferred?

I'm up for whatever. I always seem to use them in functions when you're directly assigning (`var b = kjaol;`), but not when they are declared directly in the QML (`var b: kjaol`) - no idea why I went that route

bradymadams:
> > Looks good. What is the consensus on semi-colons and QML attributes? I always use them. Looks like that's not preferred?
> 
> I'm up for whatever. I always seem to use them in functions when you're directly assigning (`var b = kjaol;`), but not when they are declared directly in the QML (`var b: kjaol`) - no idea why I went that route

Functions are actually javascript and I've been calling for semi-colons at the end of statements in JS. But I'm fine with either in the QML.

dougkenik:
> > > Looks good. What is the consensus on semi-colons and QML attributes? I always use them. Looks like that's not preferred?
> > 
> > 
> > I'm up for whatever. I always seem to use them in functions when you're directly assigning (`var b = kjaol;`), but not when they are declared directly in the QML (`var b: kjaol`) - no idea why I went that route
> 
> Functions are actually javascript and I've been calling for semi-colons at the end of statements in JS. But I'm fine with either in the QML.

Ok - I would suggest no semi-colons in QML then. I think it reads much cleaner

colman-hartinger:
> > > > Looks good. What is the consensus on semi-colons and QML attributes? I always use them. Looks like that's not preferred?
> > > 
> > > 
> > > I'm up for whatever. I always seem to use them in functions when you're directly assigning (`var b = kjaol;`), but not when they are declared directly in the QML (`var b: kjaol`) - no idea why I went that route
> > 
> > 
> > Functions are actually javascript and I've been calling for semi-colons at the end of statements in JS. But I'm fine with either in the QML.
> 
> Ok - I would suggest no semi-colons in QML then. I think it reads much cleaner

When I'm writing pure javascript I always use semi colons but for some reason it seems weird to me to use them within qml

PR #214, opened 2020-08-13
R20.1 help link

dougkenik:
Added links to access help and email help@tetonsim.com

Restyled the Table and made it easier to move it


dougkenik:
Switched

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -3,7 +3,8 @@
 from typing import Dict
 
 from PyQt5.QtWidgets import QMessageBox
-from PyQt5.QtCore import pyqtSignal
+from PyQt5.QtCore import pyqtSignal, QUrl
+from PyQt5.QtGui import QDesktopServices
```
colman-hartinger, line 8: pyqtSignal and QMessageBox are unused

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -71,6 +76,12 @@ def __init__(self):
         self._exitManager = CuraApplication.getInstance().getOnExitCallbackManager()
         self._exitManager.addCallback(self._saveOnExit)
 
+    def _openHelp(self):
+        QDesktopServices.openUrl(QUrl("https://help.tetonsim.com"))
+
+    def _contactHelp(self):
+        QDesktopServices.openUrl(QUrl("mailto:help@tetonsim.com?subject=Request for help with Smart Slice"))
+
```
colman-hartinger, line 84: Both static functions

dougkenik: @colman-hartinger - what's the reasoning? I think they should be private (hence the leading underscore) - so I'm unsure of the need for a `staticmethod` or `classmethod` decorator

colman-hartinger: from my understanding the characteristic of being static is independent of the visibility. Shouldn't make a big difference either way though

dougkenik: that's definitely true, however - an underlying property of `classmethod ` and `staticmethod` is that they are callable without an instantiation of the class itself - which I am unsure if we want in this case. What are your thoughts?

colman-hartinger: I don't see a problem with calling these methods without an instantiation of the class. In the same class _getMetaData is a private static function

dougkenik: Ok. I can see the reason for having that method static, but I think it should not be private (Must have been a previous oversight from me). If we think these routines should be accessible outside of the class without a class instantiation, I am happy to make them static

dougkenik: I moved getMetaData to a non-private function, and I kept the ones we're talking about to private and non-static. If we think I should go another route, I am happy to do so

bradymadams:
For what it's worth - private static functions make complete sense. It indicates it's a static function that you only call from within the class.

dougkenik:
done

PR #217, opened 2020-08-16
R20.1 materials warning

dougkenik:
I created some new material database lists for tested and untested materials for generic fallbacks, and retained backwards compatibility for the current database structure. If the user selects a material that is untested, a warning is issued when they enter the SS stage. If the material doesn't change, the warning will not show again if the user dismisses it. If the material changes in any stage, the warning will appear when the user enters the stage, and it will disappear if they change to a tested material.

I also changed how we are looking at active extruders for the active node, and disconnected the signal when the user changes extruders in the print settings because the active extruder for the node does not change. 

The "untested" material I added was for ABS - "Ultimaker -> ABS Black". The tested one is "Generic->ABS"


dougkenik:
@rickdalgarno - the new database structure is attached in this PR

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -110,7 +114,7 @@ def onStageSelected(self):
         application = CuraApplication.getInstance()
         controller = application.getController()
         extruderManager = application.getExtruderManager()
-        extruderManager.activeExtruderChanged.connect(self.extruderChanged)
+        # extruderManager.activeExtruderChanged.connect(self._activeExtruderChanged)
```
jake-kirkland-teton, line 117: I think you got the connection working in a different way, do we want this commented line still?

dougkenik: I actually just commented out this connection because it doesn't have an affect on the extruder assigned to the SceneNode. I'm comfortable leaving it or removing it, but I wanted to get your opinions on what I did

jake-kirkland-teton:
Actually, I'm going to have to take a step back on this. It occurred to me why that signal was setup the way it was, rather than specifically on the number 1 extruder. With the changes as they currently are, the user can select a different extruder again, which we do not want. That is something we will need to fix.

dougkenik:
> Actually, I'm going to have to take a step back on this. It occurred to me why that signal was setup the way it was, rather than specifically on the number 1 extruder. With the changes as they currently are, the user can select a different extruder again, which we do not want. That is something we will need to fix.

I'll need more elaboration here. I'm lost. As it is now, if a user selects the active extruder from 1 to 2, it actually doesn't affect the settings on the SceneNode - the user has to specifically change the active extruder on the mesh, do they not?

jake-kirkland-teton:
OK. I've lost my mind and had a bit of a flashback. The way the new prevalidate checks work takes care of what I was referring to. Ignore my previous comment. We can remove the commented line.

dougkenik:
Done - good to review

dougkenik:
> Code looks good. I can't shake the feeling that the management of material warnings in the stage feels a little misplaced. Might be more work than we want to do now, but what about putting it into the errors pop-up off the Validate button? We could switch the errors to red colors and introduce warnings (yellow). We could allow or not allow warnings to be dismissed - add a "dismiss" attribute to warnings, so they don't get shown after being dismissed.

I think you're probably right - I wasn't able to put much thought into dismissable warnings vs non-dismissible. The issue with this specific workflow is that we want to show the warning sometimes, and sometimes not, depending on user actions. I would need to think of a way to design the experience and implement hyperlinks / dismissable components in the pop-up

jake-kirkland-teton:
Other than the overarching potential rework of how the message gets displayed, I think this is almost there.

There is one last bug I found, other than that, I think it's good to go:
If you select extruder 2 on the outer wall extruder, it does not register and can give the user a weird message.
image
image

dougkenik:
> Other than the overarching potential rework of how the message gets displayed, I think this is almost there.
> 
> There is one last bug I found, other than that, I think it's good to go:
> If you select extruder 2 on the outer wall extruder, it does not register and can give the user a weird message.
> image
> image

Ok. I think this is all fixed now. I added those `EXTRUDER_KEYS` to the `ExtruderProperty ` class so it's tracked. In the future, we should add the extruder checks to pywim.

PR #233, opened 2020-08-24
moved to using Duration instead of QTime

dougkenik:
Fixed the issue of the days not displaying in the print times by switching to `UM.Duration`


SmartSlicePlugin/stage/ui/SmartSliceMain.qml
```diff
@@ -141,11 +141,16 @@ Item {
...
+                        text: smartSliceMain.proxy.resultTimeTotal.getDisplayString(UM.DurationFormat.Long)
+
                         source: UM.Theme.getIcon("clock")
                         font: UM.Theme.getFont("medium_bold")
 
                         visible: smartSliceMain.proxy.isOptimized
+
+                        onTextChanged: {
+                            console.log("PRINT TIME ", text)
+                        }
```
bradymadams, line 153: Leftover debug?

PR #248, opened 2020-08-31
fixed the lay flat mode bug

### LINKED ISSUE #244: Part Rotating After Lay Flat Mode
Take any STL, and then in Prepare, use Lay Flat Mode to rotate the part. Now move to Smart Slice and select a face for a Load / Anchor - the part "jumps" in Smart Slice because the Rotate Tool is still enabled, and Lay Flat Mode is still on

We should disable Cura tools when one of ours is activated.

dougkenik:
Fixed #244


SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -233,19 +233,27 @@ def getVisibleTools():
...
                 tool_meta_data["visible"] = not our_tools_visible
+                controller.toolEnabledChanged.emit(name, not our_tools_visible)
 
             Logger.log(
                 "d", "Visibility of <{}>: {}".format(name, tool_meta_data["visible"])
             )
 
+        # Turn off face to lay flat mode if it's on
+        if tools['RotateTool'].getSelectFaceToLayFlatMode() and our_tools_visible:
+            tools['RotateTool'].setSelectFaceToLayFlatMode(False)
```
bradymadams, line 255: Double quotes.

dougkenik:
> Looks good. Just the one style change request from me. It sucks that we have to explicitly find the tool and turn off this feature.

Agreed. I tried to just disable all of the Cura tools, but for some reason the Rotate Tool does not check if it is enabled prior to performing an operation. It's actually kind of ridiculous - Tools have an `enabled ` boolean attached to them, but I have yet to see something which actually sets it or uses it

PR #251, opened 2020-08-31
R20.1 runtime estimate

### LINKED ISSUE #240: Add runtime estimate to validation and optimization
Thor already gives back a runtime estimate. We should add the estimate to "Validating..." and "Optimizing..." so users know what to expect

jake-kirkland-teton:
This is functioning as expected. Ended up taking the simpler route of tacking it on the end of the label since the QML was being uncooperative.

Closes #240


SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -228,6 +229,13 @@ def __call__(self, job: pywim.http.thor.JobInfo) -> bool:
         elif job.status == pywim.http.thor.JobInfo.Status.running and self.connector.status not in (SmartSliceCloudStatus.BusyOptimizing, SmartSliceCloudStatus.BusyValidating):
             self.connector.status = self._previous_status
             self.connector.updateSliceWidget()
+
+        if job.status == pywim.http.thor.JobInfo.Status.running and self.connector.status is SmartSliceCloudStatus.BusyOptimizing:
+            Logger.log("d", Duration(job.runtime_estimate).getDisplayString(DurationFormat.Format.Seconds))
```
dougkenik, line 234: You still want this here?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -228,6 +229,13 @@ def __call__(self, job: pywim.http.thor.JobInfo) -> bool:
...
+            remaining_time = Duration(job.runtime_estimate)
+            if int(remaining_time.getDisplayString(DurationFormat.Format.Seconds)) > 60:
```
bradymadams, line 236: `job.runtime_estimate` is already an integer I believe, so I don't think it should be necessary to go int -> str -> int.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -228,6 +229,13 @@ def __call__(self, job: pywim.http.thor.JobInfo) -> bool:
         elif job.status == pywim.http.thor.JobInfo.Status.running and self.connector.status not in (SmartSliceCloudStatus.BusyOptimizing, SmartSliceCloudStatus.BusyValidating):
             self.connector.status = self._previous_status
             self.connector.updateSliceWidget()
+
+        if job.status == pywim.http.thor.JobInfo.Status.running and self.connector.status is SmartSliceCloudStatus.BusyOptimizing:
+            Logger.log("d", Duration(job.runtime_estimate).getDisplayString(DurationFormat.Format.Seconds))
+            remaining_time = Duration(job.runtime_estimate)
```
bradymadams, line 235: `runtime_estimate` is the overall runtime, not the remaining time. We do return the estimated remaining time, but it looks like the pywim JobInfo class needs to be updated to include it.

Do we want the overall runtime estimate or remaining runtime estimate?

dougkenik: I was just thinking of this. We want remaining time

bradymadams: I went ahead and updated pywim to include it. `JobInfo.runtime_remaining`.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: No. Good catch.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Fair enough. Though this is actually a good point, the check is no longer needed, it was in there for when we were considering displaying validate time. It has been removed.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Perfect. I have that swapped over to remaining time.

Issue #249, opened 2020-08-31
Rotating / Scaling / Transforming Mod Mesh does not invalidate


dougkenik:
https://trello.com/c/tptpqINO/57-changing-modifier-mesh-doesnt-invalidate-results

dougkenik:
I'm already working in here for a different ticket, so it makes sense just to get it done

Issue #245, opened 2020-08-29
Lose Results Warning For Extruder Changes


dougkenik:
When a user changes extruders during n run or after an optimization run for any mesh, we should prompt to lose result and switch back if they cancel - not doing that right now

dougkenik:
It might be best to put the extruder checks into the SmartSlicePropertyHandler and follow the logic there.

PR #252, opened 2020-09-01
R20.1 245

dougkenik:
Did a few things with this one:

- Moved active extruder tracking to a TrackedProperty
- Modified the ModifierMesh property tracker class to be more global for scene nodes, and included all nodes in it
- Created a class to track scene node extruder properties and added that to the Scene Node tracking
- Created a class to track scene node transform properties, and added that to the Scene Node tracking
- Moved materials warnings from the Stage to the PropertyHandler
- Migrated the PluginMetadata class to Utils

Should solve #245 and #249


dougkenik:
> Looks like almost everything is working as intended. Things rewind correctly and I like how it functions now in the prepare tab to invalidate results.
> 
> The one problem I have run into, is this changes how the Plugin Metadata works, and it no longer reads the API URL from the plugin.json file. Is this intended? I have the test environment pointed at in my json file, but the plugin is attempting to use production still.

That's not intended. Where does this happen in the code?

SmartSlicePlugin/utils/__init__.py
```diff
@@ -119,3 +119,35 @@ def angleBetweenVectors(vector1: Vector, vector2: Vector) -> float:
...
+
+        if pluginMetaData:
+            self.name = pluginMetaData.get('name', self.name)
+            self.id = pluginMetaData.get('id', self.id)
+            self.version = pluginMetaData.get('version', self.version)
+
+            apiInfo = pluginMetaData.get('smartSliceApi', None)
+
+            if apiInfo:
+                self.url = apiInfo.get('url', self.url)
```
jake-kirkland-teton, line 141: Here I believe.

jake-kirkland-teton: And then used in lines 280 - 290 in Connector

dougkenik: Ok. It was because the file location changed. I moved it back to the Extension. There was really no need to move it in the first place

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -128,6 +128,64 @@ def restore(self):
...
+                machine.setProperty(key, "value", self._specific_extruders[key])
+
+    def changed(self):
+        active_extruder_index, specific_indices = self.value()
+
+        for key, value in self._specific_extruders.items():
+            if value != specific_indices[key]:
+                return True
+
+        return active_extruder_index != self._active_extruder_index
```
colman-hartinger, line 188: I'm a little confused why there are two classes to check and restore the extruder. Could sceneNodeExtruder and activeExtruder be combined, or are they serving different purposes?

SmartSlicePlugin/utils/__init__.py
```diff
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Dict
```
colman-hartinger, line 1: unused import

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -216,16 +227,41 @@ def _onExtruderPropertyChanged(self, key: str, property_name: str):
...
         )
 
+    def _onActiveExtruderChanged(self):
+        self.confirmPendingChanges(self._active_extruder)
+
     def _onMachineChanged(self):
-        self._activeExtruder = self._activeMachineManager.activeMachine.extruderList[0]
-        self._activeExtruder.propertyChanged.connect(self._onExtruderPropertyChanged)
-        self._material = self._activeExtruder.material
+        self._activeMachineManager.activeMachine.extruderList[0].propertyChanged.connect(self._onExtruderPropertyChanged)
```
colman-hartinger, line 234: would it be possible to use the ActiveExtruderIndex instead of hardcoding in 0?

SmartSlicePlugin/SmartSliceProperty.py
dougkenik: We have a few options. The Active Extruder is global, and acts as a global property. The extruders for the scene nodes are acquired / set via the nodes themselves. We could aggregate all extruder properties under the extruder, or we could tie them to the scene nodes. Tying them to the scene nodes is a little more flexible for the future IMO

SmartSlicePlugin/SmartSlicePropertyHandler.py
dougkenik: good point - done

SmartSlicePlugin/SmartSliceProperty.py
colman-hartinger: ok I see now, probably best to leave them separate

PR #253, opened 2020-09-02
R20.1 247 250

### LINKED ISSUE #247: Empty Message Box
empty message dialog box appears on selecting "Continue" after an event to cancel changes

dougkenik:
This fixes #247

We should create a ticket to investigate how we track jobs. Right now they are placed in a container, and we append a job to the container every time we submit. I'm not convinced this is the best approach


SmartSlicePlugin/plugin.json
```diff
@@ -2,12 +2,12 @@
...
     "id" : "SmartSlicePlugin",
     "author": "Teton Simulation, Inc.",
-    "version": "20.1.32",
+    "version": "20.1.33",
     "description": "Optimize and validate slice parameters to minimize print time while meeting end-use requirements.",
     "supported_sdk_versions": ["7.0.0", "6.3.0"],
     "i18n-catalog": "smartslice",
     "smartSliceApi": {
-        "url": "https://api.smartslice.xyz:443",
+        "url": "https://test.smartslice.xyz:443",
```
bradymadams, line 10: Probably don't want to change this, right?

bradymadams:
Will merge pending approval from @colman-hartinger and we'll need to merge r20.1 into here as well.

dougkenik:
I'm looking into something really quick - please hold

dougkenik:
@bradymadams - can you take a look at that last change and tell me if it looks reasonable?

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -233,8 +233,7 @@ def __call__(self, job: pywim.http.thor.JobInfo) -> bool:
         if job.status == pywim.http.thor.JobInfo.Status.running and self.connector.status is SmartSliceCloudStatus.BusyOptimizing:
             self.connector._proxy.sliceStatus = "Optimizing...&nbsp;&nbsp;&nbsp;&nbsp;(<i>Remaining Time: {}</i>)".format(Duration(job.runtime_remaining).getDisplayString())
 
-        return False
-
+        return self.cancel
```
bradymadams, line 236: Can you `return self.connector.cloudJob.canceled`? Thus omitting the need for the added `cancel` attribute.

Issue #256, opened 2020-09-02
Load Magnitude Does not Revert on Cancel


dougkenik:
https://trello.com/c/QBWE7u6E/61-changing-load-magnitude-and-cancelling-doesnt-restore-value-in-ui

PR #259, opened 2020-09-03
R20.1 256

dougkenik:
Fix for #256


SmartSlicePlugin/select_tool/SmartSliceSelectTool.qml
```diff
@@ -252,6 +259,8 @@ Item {
                 } else {
                     flipIcon.color = UM.Theme.getColor("text_inactive");
                 }
+
+                textLoadDialogMagnitude.text = bcListForces.model.loadMagnitude=
```
bradymadams, line 263: Is that trailing `=` intentional?

bradymadams (commented):
Looks good, just the one syntax thing that I think is a typo.

This doesn't address the other problem I found with crashing when "Cancel" is clicked when the load dialog is not visible, right?

dougkenik:
> Looks good, just the one syntax thing that I think is a typo.
> 
> This doesn't address the other problem I found with crashing when "Cancel" is clicked when the load dialog is not visible, right?

It does not address that

SmartSlicePlugin/select_tool/SmartSliceSelectTool.qml
```diff
@@ -260,7 +260,7 @@ Item {
                     flipIcon.color = UM.Theme.getColor("text_inactive");
                 }
 
-                textLoadDialogMagnitude.text = bcListForces.model.loadMagnitude=
+                textLoadDialogMagnitude.text = bcListForces.model.loadMagnitude
```
bradymadams, line 263: Semi-colon would be preferred.

bradymadams (approved):
Th

PR #265, opened 2020-09-04
R20.1 260

dougkenik:
I think this is ready for a draft review while I'm out of town. I am running into an issue trying to submit a job with a modifier mesh from an optimization job (smartslice 148), but other than that it is behaving as expected. Here is what I did:

- Created a new class to track properties for all scene nodes
- Decorated scene nodes that we programatically add / remove so we don't issue a revalidation  or reoptimization warning as we change results from the table
- Fixed the issue where deleting a modifier mesh will not invalidate results
- Enhanced this by also issuing a warning when the user deletes the root node (only node) during validation or optimization, or after optimized


SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -320,6 +329,11 @@ def confirmPendingChanges(self, props, revalidationRequired=True):
 
         changes = [p.changed() for p in props]
 
+        changed = []
+        for p in props:
+            if p.changed():
+                changed.append(p)
```
bradymadams, line 335: How are `changed` and `changes` different? Is `changed` even used?

SmartSlicePlugin/plugin.json
```diff
@@ -7,7 +7,7 @@
     "supported_sdk_versions": ["7.0.0", "6.3.0"],
     "i18n-catalog": "smartslice",
     "smartSliceApi": {
-        "url": "https://api.smartslice.xyz:443",
+        "url": "https://test.smartslice.xyz:443",
```
bradymadams, line 10: I'm assuming unintentional. We need to implement a local config file so this isn't a recurring problem.

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -89,6 +92,10 @@ def __init__(self, connector):
         sel_tool.selectedFaceChanged.connect(self._faceChanged)
         sel_tool.toolPropertyChanged.connect(self._onSelectToolPropertyChanged)
         req_tool.toolPropertyChanged.connect(self._onRequirementToolPropertyChanged)
+
+        controller.getScene().getRoot().childrenChanged.connect(self._onSceneNodeChanged)
```
bradymadams, line 96: Duplicate of line 107

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -180,8 +186,9 @@ def restoreCache(self):
         """
         Restores all cached values for properties upon user cancellation
         """
+
         for p in self._properties:
-            if p.changed():
+            if p.changed() and not p != self._scene:
```
bradymadams, line 191: I believe `not p != self._scene` is equivalent to `p == self._scene`. The latter would be easier to read.

So this means we're only restoring the scene, right? Is that what we want?

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -276,6 +281,66 @@ def restore(self):
...
+            if node not in nodes:
+                if not node.getDecorator(SmartSliceRemovedDecorator):
+                    return True
+                else:
+                    _changed = True
+
+        if _changed:
+            self.cache()
+
+        return False
```
bradymadams, line 343: Not sure I understand this function. Don't we want to trigger `changed` if we detect a node has been added that we haven't seen yet. Maybe that gets done elsewhere?

bradymadams (changes requested):

SmartSlicePlugin/SmartSlicePropertyHandler.py
dougkenik: No. It was for a debug. Removed

SmartSlicePlugin/plugin.json
dougkenik: Unintentional. Agreed on creating something to make this easier

SmartSlicePlugin/SmartSlicePropertyHandler.py
dougkenik: Nope - leftover from a different way to do this

SmartSlicePlugin/SmartSliceProperty.py
dougkenik: The intent is to not return `True` if the changed node has a decorator which implies it has been programatically added / removed by Smart Slice. I'll move the functionality for caching into a different function and call it from the `onSceneNodeChanged `function

dougkenik:
We're ready for another review here

bradymadams (approved):

Issue #269, opened 2020-09-05
Changing print profile does not invalidate results


dougkenik:
https://trello.com/c/6q9GMdJt/73-changing-print-profile-does-not-invalidate-results

PR #272, opened 2020-09-07
R20.1 269

dougkenik:
Fix for #269 

It's a little weird because Cura issues a dialogue when the print profile is changed, and we need to wait for the user to take action on that dialogue before they can take action on our dialogue.


bradymadams (approved):
Looks good and behaves as expected.

PR #283, opened 2020-09-08
R20.1 print requirements ui updating bug

### LINKED ISSUE #277: Cancelling change to requirements during optimization does not change requirements back
https://trello.com/c/YTs09Auo/83-changing-smart-slice-requirements-during-an-optimize-does-not-update-the-ui-upon-cancelling

Need to follow what we did for the load magnitude and emit a signal which can be tapped into from the QML

UIUpdateBug

jake-kirkland-teton:
Closes #277


jake-kirkland-teton:
Hold on, Something isn't quite right

SmartSlicePlugin/requirements_tool/SmartSliceRequirements.qml
```diff
@@ -92,5 +101,19 @@ Item {
...
+                valueSafetyFactor.text = UM.ActiveTool.properties.getValue("TargetSafetyFactor")
+            }
+        }
+
+        Connections {
+            target: UM.Controller.activeStage.proxy
+            onUpdateTargetUi: {
+                valueMaxDeflect.text = UM.ActiveTool.properties.getValue("MaxDisplacement")
+            }
+        }
```
bradymadams, line 116: I'm not the QML expert so I may be off base here, but does it make sense to have two `Connections` that are identical except for the logic inside the function? Would the second one override the first and would it be safer to merge them?

jake-kirkland-teton: It does in fact.

bradymadams: It does in fact make sense, or overrides?

jake-kirkland-teton: Makes sense to combine them. I don't think it overrides, but I could be wrong

jake-kirkland-teton:
So the plan is to merge this as it is and count it as a currently known issue. We will have to do a deeper dive into why the values are not caching correctly when you cancel during an optimize, but the current code should be merged to address most of the scenarios in this ticket that are fixed.

dougkenik (approved):
It's running fine, except the one limitation

bradymadams (approved):

Issue #292, opened 2020-09-09
Uneeded connection in QML


nallath:
https://github.com/tetonsim/is-cura-ui/blob/14943fe08d0e0b8b0246fc09e71a1b8b22369de4/SmartSlicePlugin/stage/ui/SmartSliceMain.qml#L797-L801

This connection isn't needed. Properties are automatically handled by QML

Issue #290, opened 2020-09-09
Mismatched return type


nallath:
https://github.com/tetonsim/is-cura-ui/blob/0897c5e3414a35a939de5221060dc9ba7281e251/SmartSlicePlugin/SmartSliceJobHandler.py#L515-L542

Return on line 521 doesn't match type provided. Either change the type to Optional[Dict] or the return to `return extruder_message`

PR #293, opened 2020-09-09
Requested changes from Jaime

### LINKED ISSUE #289: Attempting to access plugin that is not required
https://github.com/tetonsim/is-cura-ui/blob/0897c5e3414a35a939de5221060dc9ba7281e251/SmartSlicePlugin/SmartSliceJobHandler.py#L277

Change to: 
threeMF_Writer = Application.getInstance().getMeshFileHandler().getWriter("3MFWriter")

You should also check if the 3MF writer is active at all. A user could have disabled the plugin.

jake-kirkland-teton:
We're getting the correct 3mf writer object, as well as ensuring the object is there.

Return type for buildExtruderMessage is now Optional[Dict]

Removed unneeded connection in Main qml

Closes #289 #290 #292


SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -274,14 +274,18 @@ def buildJobFor3mf(self, machine_name="printer") -> pywim.smartslice.job.Job:
...
-        threemf_file.close()
+            threemf_file = zipfile.ZipFile(threemf_path, 'a')
+            threemf_file.writestr('SmartSlice/job.json', job.to_json() )
+            threemf_file.close()
 
-        return True
+            return True
+
+        else:
+            return False
```
bradymadams, line 288: Take this out of the `else` so it's just the default return.

bradymadams (changes requested):
Should we take action if `write3mf` returns `False`?

SmartSlicePlugin/SmartSliceJobHandler.py
jake-kirkland-teton: Ah right. Done.

jake-kirkland-teton:
That is a good question. I'm not sure. I'm honestly not clear on how one would disable the 3mf plugin. What would we want to do if it is indeed false?

bradymadams:
> That is a good question. I'm not sure. I'm honestly not clear on how one would disable the 3mf plugin. What would we want to do if it is indeed false?

Let's show an error message that the job couldn't be submitted because the 3MFWriter plugin is disabled.

jake-kirkland-teton:
Sounds good. Done.

bradymadams:
Can you check what happens when `threeMF_Writer` is None (maybe explicitly set it to None for testing purposes). It looks like the code that calls write3mf may try to submit a `None` job (`SmartSliceCloudJob.run`), if I'm reading it correctly. Let's make sure everything is handled cleanly.

jake-kirkland-teton:
So it activates the `job.hasError()` call. And you get the following message, similar to what you would see with any other failure of the `prepareJob` method that returned None, plus the new message I added:

image

bradymadams:
Let's clean it up so the top message doesn't appear. I'm assuming the job still tried to submit, so let's avoid that since we know that we can't submit it.

dougkenik (changes requested):
UI is functioning well, besides Brady's requests. I have one request to clean up the code

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -274,14 +274,18 @@ def buildJobFor3mf(self, machine_name="printer") -> pywim.smartslice.job.Job:
     def write3mf(self, threemf_path, mesh_nodes, job: pywim.smartslice.job.Job):
 
         # Getting 3MF writer and write our file
-        threeMF_Writer = PluginRegistry.getInstance().getPluginObject("3MFWriter")
-        threeMF_Writer.write(threemf_path, mesh_nodes)
+        if Application.getInstance().getMeshFileHandler().getWriter("3MFWriter") is not None:
```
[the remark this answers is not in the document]
jake-kirkland-teton: Done.

bradymadams:
This looks much better. Nice job.

image

bradymadams (approved):
Made one minor change. Let me know if you're okay with it and I'll merge.

jake-kirkland-teton:
Looks good to me. I wasn't positive if we wanted that to be a variable or not, so if that way works for you, it works for me. Merge away

PR #295, opened 2020-09-09
R20.1 duplicate bcs

### LINKED ISSUE #286: Adding mod mesh after creating BCs results in duplicate BCs
Really easy to repro - see attached GIF. The duplicate BCs are actually created and results in a solver error if the job is submitted.

Peek 2020-09-08 19-45

jake-kirkland-teton:
The plugin metadata was not getting cleared once it was read from the project file, and it was getting duplicated when subsequent models were added. The metadata is now being cleared after it is imported.

Closes #286


dougkenik (approved):
Looks fine

bradymadams (approved):
I checked the behavior and it is no longer adding the duplicate BCs now. Glad it was an easy one!

PR #305, opened 2020-09-11
R20.1 erroneous prompt on project load

bradymadams:


SmartSlicePlugin/utils/__init__.py
```diff
@@ -79,7 +79,7 @@ def getPrintableNodes():
 def getModifierMeshes():
     return getNodes(
         lambda isSliceable, isPrinting, isSupport, isInfillMesh: \
-            isSliceable and not isPrinting and not isSupport and isInfillMesh
+            isSliceable and not isSupport and isInfillMesh
```
colman-hartinger, line 82: you can remove the `isPrinting` from the lambda

bradymadams: I don't think so. `getNodes` still sends it as an argument.

dougkenik (approved):
Needs testing on an optimization result with and without mod meshes

SmartSlicePlugin/utils/__init__.py
bradymadams: @colman-hartinger see here for what I mean

https://github.com/tetonsim/is-cura-ui/blob/07363a4240fc53e36cfac99920e09e04a27e2ec0/SmartSlicePlugin/utils/__init__.py#L66

colman-hartinger (approved):
Tested with a couple different files and everything seems to be working as expected now

jake-kirkland-teton:
I'm currently eating, but I trust you three with the testing.

jake-kirkland-teton (approved):

PR #322, opened 2020-09-15
Improved mesh error handling and SDK version updates

bradymadams:
This updates the SDK version numbers.

Also catch errors related to interactive mesh handling. UM was particularly concerned about these errors because it could be perceived by the user that it's a Cura problem and not a Smart Slice problem. So these changes will catch those exceptions and log them. In the future we can cleanup the messaging, but this will at least prevent hard crashes.


SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -152,7 +152,18 @@ def onStageSelected(self):
...
+
+            try:
+                smart_slice_node.initialize(printable_node)
+            except Exception as exc:
+                Logger.logException("e", "Unable to analyze mesh")
+                self._scene_not_ready(
+                    i18n_catalog.i18n("Smart Slice could not analyze the mesh for face selection. It may be ill-formed.")
+                )
+                if smart_slice_node:
+                    printable_node.removeChild(smart_slice_node)
```
dougkenik, line 164: Will this push the user back to the Prepare stage?

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -124,7 +124,11 @@ def _onSelectedFaceChanged(self, current_surface=None):
             if bc_node is None:
                 return
 
-            selected_face, axis = self._getSelectedTriangles(current_surface, bc_node.surface_type)
+            try:
+                selected_face, axis = self._getSelectedTriangles(current_surface, bc_node.surface_type)
+            except Exception as exc:
+                Logger.logException("e", "Unable to select face")
```
dougkenik, line 130: This is just pushing a new log statement, and allowing the user to continue selection, correct?

dougkenik (approved):
Approved, just need clarification

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
bradymadams: Correct.

SmartSlicePlugin/stage/SmartSliceStage.py
bradymadams: The above call to `_scene_not_ready` will push them back to Prepare. I purposely raised an exception when I tested this to make sure it would work and it does.

dougkenik:
Looks good

PR #323, opened 2020-09-15
Stage error message limit

### LINKED ISSUE #316: Empty build plate message when switching to Smart Slice stage should show only once
Show only once and add a timeout.

bradymadams:
Fixes #316


SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -82,11 +84,13 @@ def api(self):
...
+        if self._invalid_scene_message is None or not self._invalid_scene_message.visible:
+            title = i18n_catalog.i18n("Invalid print for Smart Slice")
 
-        Message(
-            title=title, text=text, lifetime=120, dismissable=True
-        ).show()
+            self._invalid_scene_message = Message(
+                title=title, text=text, lifetime=30, dismissable=True
+            )
+            self._invalid_scene_message.show()
```
dougkenik, line 93: I don't know how Cura messages work entirely. If the message is already shown, and we encounter a new (different) error, what happens?

bradymadams: If the previous message is still visible nothing will happen and the previous message will remain visible. I'll add a check to compare the text and if it changes I'll update the message.

colman-hartinger (approved):
Looks good to me

SmartSlicePlugin/stage/SmartSliceStage.py
bradymadams: I updated it to just hide the old one and create a new one. That way the timer resets and the text updates if it is different.

dougkenik (approved):

PR #324, opened 2020-09-15
R20.1 opt settings caching bug

### LINKED ISSUE #320: Settings change cancel on an optimized part doesn't revert to the correct setting
Reported by UM. 

The settings I had before the cancel (but after the optimize) was 4.0, not 10.0.
So slightly better reproduction steps:
1. Optimise a model
2. Check top layers value
3. Notice it's 4.0
4. Change it to something
5. Press "cancel" on the message

Expected:
1. Top layers == 4.0 (since that is what I had after optimization)

Actual:
1. Top layers  = 10 (settings prior to optimization)

jake-kirkland-teton:
Fixed the caching issue after Optimize results were loaded.

Closes #320


bradymadams (changes requested):
Looks good, but let's change the name of the signal. The signal names should represent the event that has occurred, not what you might expect a connected function to do when the signal is emitted.

Something like `optimizationResultAppliedToScene` maybe?

jake-kirkland-teton:
Sounds good. Done.

dougkenik (approved):

bradymadams (approved):

PR #325, opened 2020-09-15
Log cleanup on select tool activation

bradymadams:
1. Added `constant=True` to the appropriate pyQt properties
2. Removed unhelpful verbose debug about face select mode
3. QML was complaining about `index` being undefined. I changed it to a hardcode of 0. I didn't notice any change in the UI layout, and I imagine this results in the same layout as when index wasn't defined.


SmartSlicePlugin/select_tool/SmartSliceSelectTool.qml
```diff
@@ -678,7 +678,7 @@ Item {
                                                 };
                                                 return loadHelper.availableWidth * (index + 1) / (tickmarks.model + 1);
                                             }
-                                            x: indexHelper(index)
```
jake-kirkland-teton, line 681: This actually causes the dots on the slider to stack up on the first location. I think we need the index to be passed into the helper.

image

jake-kirkland-teton (changes requested):

SmartSlicePlugin/select_tool/SmartSliceSelectTool.qml
bradymadams: Got it. Glad you noticed that. Not sure how it works when `index` is undefined then. I'll look into it.

bradymadams: Okay, the `Repeater` actually has an index so I reverted that. But on the `Rectangle` index was not defined and that resulted in `indexHelper` not even being called, so I replaced it with just `0` explicitly. And now it looks normal and the QML warning message is gone.

jake-kirkland-teton (approved):
I believe it is working as intended now. Nicely done.

dougkenik (approved):
Functions as expected

PR #326, opened 2020-09-15
R20.1 286

dougkenik:
Fix for #286 and the issue Jaime reported when opening a saved project and Cura silently erroring out


jake-kirkland-teton (approved):
Seems like it is functioning as expected. One question about functionality that may or may not be an issue. So when a model is copied, the anchors/loads also copy, the loads just do not have an arrow anymore. But the modmesh that is created does in fact have anchors and loads of it's own. Is this enough of a problem to look into?

DuplicateModelGif

dougkenik:
> Seems like it is functioning as expected. One question about functionality that may or may not be an issue. So when a model is copied, the anchors/loads also copy, the loads just do not have an arrow anymore. But the modmesh that is created does in fact have anchors and loads of it's own. Is this enough of a problem to look into?
> 
> DuplicateModelGif

Yeah, we should just remove those from the modifier mesh

dougkenik:
Ok - we're good now

bradymadams (approved):

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -179,6 +179,15 @@ def onStageSelected(self):
...
             mesh.setSelectable(False)
 
+            # Remove any HighlightFace if they exist
+            for node in mesh.getChildren():
+                if isinstance(node, SmartSliceScene.HighlightFace):
+                    mesh.removeChild(node)
+                elif isinstance(node, SmartSliceScene.Root):
+                    mesh.removeChild(node)
+
+            a = 1
```
bradymadams, line 189: Wait, what's `a`?

jake-kirkland-teton (approved):

Issue #321, opened 2020-09-15
Mark changed settings as "changed"


bradymadams:
Reported by UM. There may be something we can do here, I think this might be related to changing the property "state", but I'm not sure. It may also be out of our hands:

Steps:
1. Optimise a print

Expected:
1. Settings are added and a "changed settings" is visible in the profile dropdown. 

Actual:
1. Changed settings are not registered as such
Notes; I'm not sure where this bug lies. It could be that it's a Cura issue. The issue is that the star icon doesn't appear.

PR #328, opened 2020-09-15
Partial Fix for #321

jake-kirkland-teton:


bradymadams:
Now I remember this - I don't think we can use `set_from_cache=False` because, if I recall correctly, this triggers property change signals to emit. We then get those emissions and think the property was changed by the user. That was a long time ago though, so maybe our property tracking doesn't suffer from that problem anymore.

dougkenik:
Yeah, you're right. The only way I see around this is to also track the `state ` variable for all of the extruder settings, and then we could maybe use that to determine if we cache the properties or not

dougkenik:
actually - @bradymadams, the way Jake implemented it works really well. The only thing we can't do is this:

1. Optimization, or validating
2.  Change the print profile
3. Select "Discard" when it asks if you want to Keep or Discard the changes
4. Select "Cancel" from our prompt
5. Everything reverts fine, but the Star icon next to the print profile doesn't come back (it should)

colman-hartinger (commented):
All the code looks good to me. Will run everything tomorrow morning when I'm back in the office

colman-hartinger (approved):
Just ran some optimization jobs and the changed setting is now visible. Didn't seem to effect anything else

bradymadams (approved):
@dougkenik sorry, forgot about this one. Do you want to change it to merge to r20.2?

dougkenik:
> @dougkenik sorry, forgot about this one. Do you want to change it to merge to r20.2?

I'm open to putting it into the r20.1 branch that we just made the fix for the Cura crash or put it into r20.2. What are your thoughts?

bradymadams:
> > @dougkenik sorry, forgot about this one. Do you want to change it to merge to r20.2?
> 
> I'm open to putting it into the r20.1 branch that we just made the fix for the Cura crash or put it into r20.2. What are your thoughts?

Unless, there's a reason we need to create another r20.1 release it should go into r20.2.

jake-kirkland-teton:
Want me to merge it into 20.2 then?

dougkenik:
Yeah

jake-kirkland-teton:
Should be clear to merge.

bradymadams (changes requested):
Empty lines got added to the JSONs. Let's remove those and then I can merge it.

jake-kirkland-teton:
> 
> 
> Empty lines got added to the JSONs. Let's remove those and then I can merge it.

So they did. Fixed.

bradymadams (approved):

PR #331, opened 2020-09-22
R20.2 303

### LINKED ISSUE #303: Properties Cached when Switching To Smart Slice Stage
If you change a property to invalidate results in the Prepare or Preview stage, move to the Smart Slice stage, and Cancel the changes, the properties don't revert.

Due to properties being cached when switching to the Smart Slice stage

https://trello.com/c/3VYsNhla/93-changed-print-setting-not-reverted-upon-selecting-cancel-on-smart-slice-warning-dialog

### LINKED ISSUE #329: 3MF Loading Face Highlighting bug
If you load in a 3MF with Optimization results, the boundary conditions do not highlight correctly. We either need to cause the print to go back to the prepare tab first when loaded, or figure out how to show the highlighting correctly.

dougkenik:
Rundown of changes. I didn't mean to do this all in one branch, but I started unraveling a thread

1. Fixed #303 
2. Fixed #329
3. If we open a new workspace when an optimization is already open, the prompt to invalidate results does not appear now
4. I found a bug introduced by the fix for #305, and realized it did not actually solve the problem. I found the root issue and fixed it.


colman-hartinger (approved):
All the code seems good and when I ran everything seems to solve the mentioned bugs

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -9,6 +9,10 @@
 from UM.Application import Application
 from UM.Extension import Extension
 from UM.PluginRegistry import PluginRegistry
+from UM.Job import Job
+from UM.JobQueue import JobQueue
```
bradymadams, line 13: Unused.

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -252,6 +265,12 @@ def _reset(self, *args):
         if len(getPrintableNodes()) == 0:
             self._storage.getPluginMetadata(self.metadata.id).clear()
 
+    def _workspaceLoading(self, job: Job = None):
```
bradymadams, line 268: Would `job` ever actually be `None` here? If not, let's remove the check on `if job`, otherwise if it really can than the type should be `Optional[Job]`.

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -74,6 +79,9 @@ def __init__(self):
         self._exitManager = CuraApplication.getInstance().getOnExitCallbackManager()
         self._exitManager.addCallback(self._saveOnExit)
 
+    def _onEngineCreated(self):
+        Application.getInstance()._job_queue.jobStarted.connect(self._workspaceLoading)
```
bradymadams, line 83: This seems dicey. I don't like the idea that we have to connect to a signal that will emit at the start of every job. I'm assuming connecting to `ReadFileJob.finished` doesn't have the same effect?

jake-kirkland-teton (approved):
Minus the one comment, everything looks good.

bradymadams (changes requested):

SmartSlicePlugin/SmartSliceExtension.py
bradymadams: And on second glance, if you have to connect to this signal wouldn't it be better to connect to `JobQueue.jobStarted`, rather than accessing the private attribute in the application? Although, I'll admit - I'm loose on the details of why Signals are created at the class level and what effect that has.

SmartSlicePlugin/SmartSliceExtension.py
dougkenik: No, the job is checked if it's none in the JobQueue before the signal is emitted

SmartSlicePlugin/SmartSliceExtension.py
dougkenik: I tried to connect to `JobQueue.jobStarted` at first, but it wouldn't work. I am very confused as to how that is possible, but it is the case.

The whole thing is dicey. I hate looking into private variables and connecting to signals which absolutely no other routine ever connects to. Ideally, Cura would emit a signal upon workspace reading beginning and ending, but we only get the signal at the end - which we cannot connect to because the scene has already been changed.

I'll look into connecting to `ReadFileJob.finished`, but I believe I decided against it for some reason.

dougkenik: Ok - so I cannot connect to `ReadFileJob.finished` because I would need to connect to the actual job itself, which I do not have access to because it's managed by `WorkspaceFileHandler`. I can get the instance of `WorkspaceFileHandler`, from `CuraApplication`, but there is no way to get the actual Job which is reading the file

bradymadams (approved):

Issue #297, opened 2020-09-10
Modifier mesh location not read in correctly


dougkenik:
Modifier mesh was generated with SmartSlice, the position is below.

If I revalidate with the modifier mesh and save it, then open the file, the modifier mesh is not positioned on the build plate correctly:

[VALIDATED_TEST.3mf.zip](https://github.com/tetonsim/is-cura-ui/files/5209054/VALIDATED_TEST.3mf.zip)
![saved_position](https://user-images.githubusercontent.com/23110600/92927447-da6a6200-f3fa-11ea-95b1-0ef6dde0c7e1.png)
![opened_position](https://user-images.githubusercontent.com/23110600/92927452-db9b8f00-f3fa-11ea-9cce-767c3923958d.png)

bradymadams:
@dougkenik is it only with manually added mod meshes or does it also occur with mod meshes from optimization?

This seems strange since we're not managing mod meshes during the saving and loading, right?

dougkenik:
This happened after an optimization. I modified the modifier mesh density, then revalidated, and saved. When I open it up again, the mesh is out of place.

We do not handle their position, but we do create a ZOffset decorator for nodes from the optimization. It is likely happening from that

bradymadams:
Okay, let's see what it takes to fix it.

dougkenik:
I think what is happening is the plugin metadata isn't being cleared between saves. Verifying now - will take a while though

dougkenik:
@rickdalgarno - let's document this as a known issue.

Here are the steps:

- User optimizes 
- Selects a result with a modifier mesh
- Changes anything to invalidate the results and continues
- Save the project
- Open the project

The modifier mesh will be positioned incorrectly in the Z location (looks like by about 1/2 of the modifier mesh)

dougkenik:
I think the best option here would be to save the modifier meshes to our pywim.results object and then use the functions in SmartSliceProxy to remove any mod meshes and re-insert ours

bradymadams:
> I think the best option here would be to save the modifier meshes to our pywim.results object and then use the functions in SmartSliceProxy to remove any mod meshes and re-insert ours

I agree that seems like a viable fix, but we should probably take a little time to understand more of what's going on with the transformations and the 3mfs before we commit to that.

PR #333, opened 2020-09-22
fixed the z offset bug for saved validations from optimization

dougkenik:
Fix for #297. To test, you'll need to generate an optimized result with a mod mesh, invalidate a result with a mod mesh, save that file. You can re-open that file and see the mesh is not under the build plate. I have attached an optimized model you can try out.

[OPTIMIZED_TEST.zip](https://github.com/tetonsim/is-cura-ui/files/5264263/OPTIMIZED_TEST.zip)


SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -813,9 +814,7 @@ def updateSceneFromOptimizationResult(self, analysis: pywim.smartslice.result.An
             modifier_mesh_node.setCalculateBoundingBox(True)
 
             # Use the data from the SmartSlice engine to translate / rotate / scale the mod mesh
-            parent_transformation = our_only_node.getLocalTransformation()
-            modifier_mesh_transform_matrix = parent_transformation.multiply(Matrix(modifier_mesh.transform))
-            modifier_mesh_node.setTransformation(modifier_mesh_transform_matrix)
+            modifier_mesh_node.setTransformation(Matrix(modifier_mesh.transform))
```
bradymadams, line 817: I seem to recall this is what it used to be, but it was changed a couple months ago to account for some other bug? Maybe I'm not recalling correctly, thought.

dougkenik: I don't think it was changed, I moved "up" in the function so I could fix some other things. The base of the changes are that the modifier mesh needs to have the printable node as its parent instead of `Root`. Once that happens, the transformation needs to reference `parent ` instead of `world`.

jake-kirkland-teton (approved):
I think this looks alright if Brady's concerns are covered. Seems to work as intended.

colman-hartinger (approved):

bradymadams (approved):

PR #334, opened 2020-09-23
R20.2 280

### LINKED ISSUE #280: Track position of meshes
Need to track the position of mod meshes and the other mesh in relation to one another, if status is optimized and a mod mesh exists on the build plate:

MovingModelMeshBug

dougkenik:
Ok, this fixes #280 and needs to be validated a few different ways:

1. If a user brings in a normal and modifier mesh, runs a validation, and moves either mesh, it should invalidate results
2. Same with above but with opening a validation that has been ran and saved
2. If we run an optimization and insert a modifier mesh. If the user moves the normal mesh, the modifier mesh should move with it and not invalidate results. If the user moves the modifier mesh it should invalidate the results


jake-kirkland-teton:
So if you move something during a validate, then cancel it, it causes an error in the 3MF that is generated from there on out.

CancelChangesGif

dougkenik:
> So if you move something during a validate, then cancel it, it causes an error in the 3MF that is generated from there on out.
> 
> CancelChangesGif

Save the 3mf and post it here so someone can look into why it's causing an issue

jake-kirkland-teton:
[UMS5_Flat_Bracket_modmesh_bug.zip](https://github.com/tetonsim/is-cura-ui/files/5272213/UMS5_Flat_Bracket_modmesh_bug.zip)

dougkenik:
> [UMS5_Flat_Bracket_modmesh_bug.zip](https://github.com/tetonsim/is-cura-ui/files/5272213/UMS5_Flat_Bracket_modmesh_bug.zip)

@bradymadams - can you look at that when you're free and let me know if anything strange stands out

bradymadams:
Our 3MF reader chokes on the empty arrays on the mod mesh settings. I opened an issue in pywim: https://github.com/tetonsim/pywim/issues/67. If I remove those settings it runs for me.

There are a lot of settings on that mod mesh. I'd remove any that aren't necessary - our 3MF reader struggles with those (also related to https://github.com/tetonsim/pywim/issues/63).

image

jake-kirkland-teton:
So if the bug that was created is not problematic for closing this ticket, and we want to handle the bug that was created in a separate ticket, I approve.

jake-kirkland-teton (approved):

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -273,31 +275,60 @@ def changed(self) -> bool:
...
         self._orientation = None
+        self._position = None
+        self._z_offset = None
 
     def value(self):
         if self._node:
-            return self._node.getScale(), self._node.getOrientation()
-        return None, None
+            return self._node.getScale(), self._node.getOrientation(), self._node.getPosition(), self._node.getDecorator(ZOffsetDecorator), self.findIntersectingNodes()
+        return None, None, None, None, []
```
colman-hartinger, line 289: I think the changes to this class may be cleaner if all the transform objects were added to an array

colman-hartinger (approved):
The code runs fine and couldn't find any bugs when, I have my  comment above a bout possibly putting all the transform objects into an array but I'm not sure if that's a viable option or not

colman-hartinger:
> 
>     3. If we run an optimization and insert a modifier mesh. If the user moves the normal mesh, the modifier mesh should move with it and not invalidate results. If the user moves the modifier mesh it should invalidate the results

I remember when I was working on the mod mesh validation ticket that the backend had some problems validating parts with mod meshes if the mod mesh is defined as a child of the printable node. I'm running on optimization right now so hold of on merging the code for a sec @bradymadams

colman-hartinger (changes requested):
@dougkenik I would revert back to not having the modifier meshes as a child node of the printable mesh since the backend can't properly validate that.

dougkenik:
> @dougkenik I would revert back to not having the modifier meshes as a child node of the printable mesh since the backend can't properly validate that.

We cannot do that because the position of the meshes will get messed up on project open. What is the message from the back end?

colman-hartinger:
> We cannot do that because the position of the meshes will get messed up on project open. What is the message from the back end?

It's not that you get an error from the backend its that the transformation gets messed up during validation. I didn't save the optimized job I just ran but I'll run another to post for example.

The problem is that the results will be vastly different from the optimized results. So if you take a part with a modifier mesh that  you just got back from an optimization and then go back to validate it with the exact same loading and anchors you will get very different results when the modifier mesh is defined as a child of the printable node. In my test using a brake lever with a modifier mesh generated from an optimization the results were a fos: 1.6 and max displacement: 3.45 under the same loading and anchors in validation the results went up to fos: 2.3 and max displacement: 4.5

colman-hartinger:
Sorry I shoudl've realized this in your previous PR about the z-offset but the problem didn't click until I read "If the user moves the normal mesh, the modifier mesh should move with it "

dougkenik:
> > We cannot do that because the position of the meshes will get messed up on project open. What is the message from the back end?
> 
> It's not that you get an error from the backend its that the transformation gets messed up during validation. I didn't save the optimized job I just ran but I'll run another to post for example.
> 
> The problem is that the results will be vastly different from the optimized results. So if you take a part with a modifier mesh that you just got back from an optimization and then go back to validate it with the exact same loading and anchors you will get very different results when the modifier mesh is defined as a child of the printable node. In my test using a brake lever with a modifier mesh generated from an optimization the results were a fos: 1.6 and max displacement: 3.45 under the same loading and anchors in validation the results went up to fos: 2.3 and max displacement: 4.5

Ok, well it's likely because our 3MF reader is not recognizing the node is a child of the parent, and not applying the transformation correctly. We might want to consider updating our 3MF reader to handle this. I could see this just happening from other potential workflows. @bradymadams  - thoughts?

bradymadams:
Can someone provide me with an example 3MF that doesn't get processed correctly on the backend?

colman-hartinger:
@bradymadams here you go. 
[UMS5_simple_lever.zip](https://github.com/tetonsim/is-cura-ui/files/5285525/UMS5_simple_lever.zip)

bradymadams:
This probably isn't surprising, but in the 3MF in this case the mod mesh gets grouped as a component under the normal mesh. We don't currently read in the `component` tags, so in this case we would completely omit the mod mesh from the smart slice analysis. I created an issue to support this: https://github.com/tetonsim/py3mf/issues/2

```xml
        <resources>
                <object id="1" name="simple_lever.stl" type="model">
			<components>
				<component objectid="2" transform="1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0" />
			</components>
		</object>
		<object id="2" name="SmartSliceMeshModifier" type="model">
                    ...
		</object>
	</resources>
	<build>
		<item objectid="1" transform="1.0 0.0 0.0 0.0 0.0 1.0 0.0 -1.0 0.0 165.0 120.0 6.0" />
	</build>
```

bradymadams:
This PR should fix the problem with grouping the mod mesh: https://github.com/tetonsim/pywim/pull/69

jake-kirkland-teton (approved):

colman-hartinger (approved):
Just tested out and mod meshes are validating with same results as optimization

bradymadams (approved):

PR #335, opened 2020-09-23
fixed the enabled tool button bug

### LINKED ISSUE #307: Sometimes the Use Case button is disabled
Originally reported by @dougkenik 

> I've seen this happen a few times, but it's hard to reproduce. Sometimes the Use Case button becomes disabled, and you can't click on it. The way to get it back is to deselect the model, and then reselect it

use-case-button-disabled

dougkenik:
Fixes #307 

This was happening because we were manipulating the `_enabled` boolean in the `event` method to turn off face selection if the user switched to Requirements (there was a bug where a user could select a face for a load or BC if Requirements was enabled), and the button icon is a UM.`RecolorImage`, where the enabled state is dependent on that specific boolean. If we used the Translate Tool and flipped to the Smart Slice tab, the button would become diabled.

I created a different variable inside the `SelectTool` to indicate if face selection should be enabled, and set it based on the `event` method.


jake-kirkland-teton (approved):
Looks like it is working as intended. I tried a few workflows and loading a few different saves, the button is enabled as it should be.

bradymadams (approved):

Issue #317, opened 2020-09-15
Unexpected error log "No view named None found"


bradymadams:
1. Don't have a model on the build plate
2. Try to switch to smart slice stage
Expected;
1. No error log
Actual:
```
2020-09-15 10:55:41,365 - DEBUG - [MainThread] cura.AutoSave._onTimeout [60]: Autosaving preferences, instances and profiles
2020-09-15 10:55:41,748 - DEBUG - [MainThread] UM.Controller.setActiveStage [180]: Setting active stage to SmartSlicePlugin
2020-09-15 10:55:41,760 - DEBUG - [MainThread] UM.Controller.setActiveStage [180]: Setting active stage to PrepareStage
2020-09-15 10:55:41,762 - DEBUG - [MainThread] UM.Controller.setActiveView [108]: Setting active view to None
2020-09-15 10:55:41,765 - ERROR - [MainThread] UM.Controller.setActiveView [120]: No view named None found
```

Look into why the `ERROR` is being reported and see if we can eliminate it.

Issue #312, opened 2020-09-11
Modifier mesh results ambiguous w.r.t. location


bradymadams:
Originally reported by @jlselden

> With a saved validation 3mf, the presentation of the models can be different if "Automatically drop models to the build plate" is checked when the file is opened as a project. The results are the same (see pics), but it is very non-intuitive.

![modmeshpos1](https://user-images.githubusercontent.com/23110600/92937741-f412a600-f408-11ea-9ae9-9a184b801422.png)
![modmeshpos2](https://user-images.githubusercontent.com/23110600/92937749-f5dc6980-f408-11ea-9226-0545bb7d723e.png)

bradymadams:
@jlselden do you have that 3MF handy to attach here?

jlselden:
This one works. 

[UMS5_dogbone_save4.zip](https://github.com/tetonsim/is-cura-ui/files/5241874/UMS5_dogbone_save4.zip)

dougkenik:
@jlselden can you tell me the exact steps you took get the Validation file saved with the mesh in that location? Was it an optimization at first, or did you get the mesh there somehow, etc..

jlselden:
I loaded the dogbone model. Then loaded the modifer mesh model (which I had lying around). Then performed a validation on it and saved that.

It was not a result from an optimization run. I was just playing with adding modifier meshes, validating, saving, and opening.

dougkenik:
> I loaded the dogbone model. Then loaded the modifer mesh model (which I had lying around). Then performed a validation on it and saved that.
> 
> It was not a result from an optimization run. I was just playing with adding modifier meshes, validating, saving, and opening.

Ok, so when you brought in the modifier mesh model, did you move it to the desired location before validating? If so, was the Drop on Build Plate setting enabled?

jlselden:
No, it wasn't. I had to turn that off in order to place the modifier mesh off the build plate.

dougkenik:
> No, it wasn't. I had to turn that off in order to place the modifier mesh off the build plate.

Was it turned back on when you brought the models back in?

jlselden:
I varied that. If it was off, everything was fine. If it was on, then the issue happened.

dougkenik:
Ok, thanks. So it only happens if the setting is off, the project is saved, then the setting is on, and the project is opened

jlselden:
Yes, that is what I saw.

Issue #304, opened 2020-09-10
Job Status Validating/Optimizing is saved with project


jake-kirkland-teton:
https://trello.com/c/CFWQDaCG/94-button-state-is-saved-in-3mf

Issue #266, opened 2020-09-04
The save warning dialog always comes up in the corner of the screen


bradymadams:
Would be a more common UX if it popped up in the center of the screen.

![image](https://user-images.githubusercontent.com/23110600/92280535-d5675900-eeb6-11ea-9630-379a7994bdae.png)

dougkenik:
@bradymadams - I assume the About Box does this for you too? The x and y positions for these QML Dialogs are not set. However, this is the only time I have not seen them in the center of the screen

bradymadams:
> @bradymadams - I assume the About Box does this for you too? The x and y positions for these QML Dialogs are not set. However, this is the only time I have not seen them in the center of the screen

Actually, the about dialog comes up dead center in the screen for me.

PR #338, opened 2020-09-28
R20.2 312

dougkenik:
Fix for #312, #317, #266, and #304 

For #312 - I decided the best way to solve this is to make any modifier meshes which intersect the normal mesh a child of the normal mesh when the project is saved.


jake-kirkland-teton (approved):
Changes look good. The bugs appear fixed, as long as the save dialog appears correctly for Brady.

SmartSlicePlugin/SmartSliceSavePrompt.qml
```diff
@@ -15,15 +15,18 @@ UM.Dialog {
...
 
     minimumWidth: width;
     maximumWidth: width;
 
     minimumHeight: height;
     maximumHeight: height;
 
+    x: 0.5 *(CuraApplication.appWidth() - width)
+    y: 0.5 *(CuraApplication.appHeight() - height)
+
```
colman-hartinger, line 29: Add space after *

colman-hartinger (changes requested):
Code looks good and bug seems fixed. Just one quick styling change and should be good to go

colman-hartinger (approved):

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -285,7 +286,7 @@ def __init__(self, node: CuraSceneNode = None):
 
     def value(self):
         if self._node:
-            return self._node.getScale(), self._node.getOrientation(), self._node.getPosition(), self._node.getDecorator(ZOffsetDecorator), self.findIntersectingNodes()
+            return self._node.getScale(), self._node.getOrientation(), self._node.getPosition(), self._node.getDecorator(ZOffsetDecorator), intersectingNodes(self._node)
```
bradymadams, line 289: Break up long line

bradymadams (changes requested):
The save dialog looks good on my machine now.

I have some questions about checking the intersecting nodes and setting parents.

1. Does this have the potential to set a parent, or set a node's parent to a different node when we wouldn't want to? Mainly I'm thinking about someone using Cura for a non Smart Slice workflow - is it possible we'd manipulate the scene in a way they wouldn't want us to when they save the 3MF?
2. Will this be a non-issue for any 3MF saved with the fix from https://github.com/tetonsim/is-cura-ui/pull/334

dougkenik:
> The save dialog looks good on my machine now.
> 
> I have some questions about checking the intersecting nodes and setting parents.
> 
> 1. Does this have the potential to set a parent, or set a node's parent to a different node when we wouldn't want to? Mainly I'm thinking about someone using Cura for a non Smart Slice workflow - is it possible we'd manipulate the scene in a way they wouldn't want us to when they save the 3MF?
> 2. Will this be a non-issue for any 3MF saved with the fix from #334

1. Yes, if the Plugin is installed, and a user saves, then we could set or change the parent. I went back and forth on this, and it seems like the best option to solve this bug (I have no idea how to do it otherwise). I also have no clue how a SceneNode can get set as a parent or child in any other Cura workflow (besides, maybe, grouping)
2. No - this is a very different problem, and happens on project open for a modifier mesh which does not have a parent. The modifier mesh is always dropped to the build plate if it does not have a parent. For optimizations, we set a parent for the mod mesh, so it gets read in at the correct location and will not get dropped to the build plate (which is why I chose to go this route)

bradymadams:
Is the issue with manually added modifier meshes only? If so, maybe we can check for modifier meshes when a validation starts and if they're parent is the root reset it to the one printable node. Would that work?

dougkenik:
> Is the issue with manually added modifier meshes only? If so, maybe we can check for modifier meshes when a validation starts and if their parent is the root reset it to the one printable node. Would that work?

This is the case. We could do that, but then there is an issue that the modifier meshes would not move independently like when they were brought in. I chose to do this on project save because then the UI appears to be consistent if the user is iterating on moving modifier meshes. Once the mod mesh gets a parent, the mod mesh is moved automatically when the parent is moved, but the mod mesh can be moved without the parent being moved. The other issue is that if a user sets up a validation, does not run, and saves, the modifier mesh could be dropped to the build plate when they open again

bradymadams:
Okay, well, I really don't like that this changes the scene of non Smart Slice workflows. I think we need to figure out a different approach.

dougkenik:
> Okay, well, I really don't like that this changes the scene of non Smart Slice workflows. I think we need to figure out a different approach.

Ok... Any way we automatically solve this problem will result in the scene changing. Even if we find a way to not set the parent of the modifier mesh, on project open we would need to adjust the Z-location of the modifier mesh (which I do not think is feasible from what I've seen when we open 3MFs). We could set the parent of the modifier mesh when the user enters the stage or has any interaction in the stage - that would indicate the user has "intent" to use Smart Slice, and we would need to remember the position of the modifier mesh.

The other way would be to add the modifier mesh to our Smart Slice metadata in the Cura project and re-insert it when they enter the stage... it's not elegant though, and we would need to remove it and insert the Cura one if they leave the stage

bradymadams:
> We could set the parent of the modifier mesh when the user enters the stage or has any interaction in the stage - that would indicate the user has "intent" to use Smart Slice, and we would need to remember the position of the modifier mesh.

I think this is definitely better. And it would remove all ambiguity if we prompted the user before doing changing the node's parent.

dougkenik:
At this point, we might as well just prompt the user on project save. If they decline, it's on them

bradymadams:
> At this point, we might as well just prompt the user on project save. If they decline, it's on them

I guess I'm fine with either since they both require user confirmation before changing the scene.

dougkenik:
> > At this point, we might as well just prompt the user on project save. If they decline, it's on them
> 
> I guess I'm fine with either since they both require user confirmation before changing the scene.

Ok, I moved all of the logic to `onStageSelected` and will issue a dismissible message when we change the parent of a mod mesh.

jake-kirkland-teton (approved):
Changes look good. The bugs appear fixed, as long as the save dialog appears correctly for Brady.

bradymadams (approved):
Will merge pending approval from @colman-hartinger

colman-hartinger (approved):

Issue #319, opened 2020-09-15
Set integer settings as integers


bradymadams:
Reported by UM that we are setting some integer settings as floats. We didn't get any specifics on which ones, so check the settings we modify and make sure any that should be integers stay integers when we change them.

PR #339, opened 2020-09-29
set properties depending on type

dougkenik:
Fix for #319


bradymadams (commented):
It would probably be more reliable to get the type from the container stack since this is what it _should_ be, rather than what the current value type is.

image

dougkenik:
> It would probably be more reliable to get the type from the container stack since this is what it _should_ be, rather than what the current value type is.
> 
> image

that's fine. Is there an easy way to convert from a string representation of the type to the type itself? If not, I can create a function to do it.

bradymadams:
[Looks likes this might be a feasible way to do it](https://stackoverflow.com/questions/11775460/lexical-cast-from-string-to-type), but I can't really endorse it as I've never used the `pydoc` module. If it's just a few basic types I think a simple `if/elif` block is fine, or a dictionary lookup.

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -792,7 +801,10 @@ def updateSceneFromOptimizationResult(self, analysis: pywim.smartslice.result.An
 
             for key, value in extruder_dict.items():
                 if value is not None:
-                    active_extruder.setProperty(key, "value", value, set_from_cache=True)
+                    property_type = type_map[active_extruder.getProperty(key, "type")]
+                    active_extruder.setProperty(
+                        key, "value", pywim.smartslice.val.convert(property_type, value), set_from_cache=True
```
bradymadams, line 806: Is there a need for this call to a `pywim.smartslice.val.convert`? What's wrong with just `property_type(value)`?

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -792,7 +801,10 @@ def updateSceneFromOptimizationResult(self, analysis: pywim.smartslice.result.An
 
             for key, value in extruder_dict.items():
                 if value is not None:
-                    active_extruder.setProperty(key, "value", value, set_from_cache=True)
+                    property_type = type_map[active_extruder.getProperty(key, "type")]
```
bradymadams, line 804: This will crash Cura if the return from `getProperty` isn't in `type_map`. You should either check for the existence of the key in `type_map` or use `type_map.get` with a default value.

bradymadams (changes requested):

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -792,8 +801,12 @@ def updateSceneFromOptimizationResult(self, analysis: pywim.smartslice.result.An
 
             for key, value in extruder_dict.items():
                 if value is not None:
-                    active_extruder.setProperty(key, "value", value, set_from_cache=True)
-                    active_extruder.setProperty(key, "state", InstanceState.User, set_from_cache=True)
+                    property_type = type_map.get(active_extruder.getProperty(key, "type"))
```
bradymadams, line 804: Now if the return of `getProperty` doesn't exist in `type_map` you'll get `None` back since you didn't specify a default, which will cause a crash on line 807.

bradymadams (changes requested):

SmartSlicePlugin/SmartSliceCloudProxy.py
dougkenik: How is that, I'm checking to see if `property_type` is `None` on line 805?

bradymadams: Missed that, sorry.

bradymadams (approved):

jake-kirkland-teton (approved):

colman-hartinger (approved):

PR #341, opened 2020-10-05
R20.2 api unittests

colman-hartinger:


colman-hartinger:
More or less just a rough draft and some of the naming could obviously be better but everything is working so I'm opening this up for comments. 

I also think it could be useful to add pylint or a [codeQL](https://github.com/github/codeql-action) check to make more use of the automated tests but that should probably be split off into another ticket

.github/workflows/smartslice-test.yml
```diff
@@ -0,0 +1,16 @@
...
+  push:
+
+jobs:
+  build:
+    container: tetonsim/smartslicetests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout SmartSlice
+      uses: actions/checkout@v2
+    - name: test 1
```
bradymadams, line 13: Rename to "Unit Tests" or something else more descriptive.

run_in_docker.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# UNCOMMENT FOR A SPECIFIC BRANCH OF PYWIM/PY3MF
+#git clone https://github.com/tetonsim/pywim.git
+#git clone https://github.com/tetonsim/py3mf.git
+#mv pywim/pywim SmartSlicePlugin/3rd-party/cpython-common
+#mv py3mf/threemf SmartSlicePlugin/3rd-party/cpython-common
+
+git clone -b 4.6 --single-branch  https://github.com/Ultimaker/Cura.git
```
bradymadams, line 9: We will probably want to parameterize the branch name and use that as an argument into this script that we get from a [build matrix in the github action](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix).

Unnecessary extra space after the `--single-branch` argument.

run_in_docker.sh
```diff
@@ -0,0 +1,21 @@
...
+mv SmartSlicePlugin Cura/plugins
+
+python3 -m pip install PyQt5==5.10  teton-3mf teton-pywim
```
bradymadams, line 12: Unnecessary extra space after the `PyQt` argument.

.github/workflows/smartslice-test.yml
```diff
@@ -0,0 +1,16 @@
+name: Smart Slice Tests
+
+on:
+  push:
+
+jobs:
+  build:
+    container: tetonsim/smartslicetests
```
bradymadams, line 8: Should we use a specific tag here (and not `latest`) so that any updates to the image can be managed to not break old branch tests? It's probably not super important, but I can imagine scenarios where we're working on a new release but need to hot patch an old release and we might need to use different images between the two development paths.

SmartSlicePlugin/tests/run.py
```diff
@@ -0,0 +1,81 @@
...
+    def unittest_wait_and_run():
+        while wait_for_app:
+            time.sleep(0.1)
+
+        plugins = PluginRegistry.getInstance()
+        prog = unittest.main(exit=False, verbosity=2, argv=[sys.argv[0]])
+
+        app.closeApplication()
+
+        # TODO use prog.result to take appropriate action based off the results
```
bradymadams, line 72: Can the `TODO` be removed?

SmartSlicePlugin/tests/test_API.py
```diff
@@ -0,0 +1,263 @@
...
+from typing import Callable
+
+from UM.PluginRegistry import PluginRegistry
+from UM.Application import Application
+
+import pywim
+
+from SmartSliceTestCase import _SmartSliceTestCase
+
+class MockJob():
```
bradymadams, line 13: The Mock classes are probably fine here for now, but eventually these will need to be pulled out to a location where they can be used across more TestCases. It should be obvious, I suppose, as you guys expand out the unit tests.

SmartSlicePlugin/tests/test_API.py
```diff
@@ -0,0 +1,263 @@
...
+        self._api._client._subscription = None
+        self._api._client._job_test = "finished"
+
+        self._api._app_preferences.removePreference(self._api._username_preference)
+        self._api._app_preferences.addPreference(self._api._username_preference, "old@email.com")
+
+    def tearDown(self):
+        pass
+
+    def test_00_check_token_create(self):
```
bradymadams, line 149: Let's remove the numbers from the test names. It will only cause a headache when you want to add a new test in the middle and now you have to change all of the test names.

bradymadams (changes requested):
Looks like a pretty foundation to start with, guys. Just a handful of straight forward change requests.

The "DOCKERFILE" should be spelled "Dockerfile".

Let's rename "run_in_docker.sh" to something more descriptive of what it's doing, such as "run_unit_tests.sh"

It's hard to parse out our test logs from all of the info being dumped to stdout/stderr. Can we print a summary at the end of the run of our unit test results?

SmartSlicePlugin/tests/run.py
jake-kirkland-teton: I think it's pretty safe to remove the TODO. Done.

SmartSlicePlugin/tests/test_API.py
jake-kirkland-teton: I moved them to MockClasses.py, might as well move them early.

SmartSlicePlugin/tests/test_API.py
jake-kirkland-teton: Done.

SmartSlicePlugin/tests/test_API.py
bradymadams: If it's alright with you guys I'd prefer to use a more standard Python naming approach under the tests module/classes. I think `mock.py` would suffice.

jake-kirkland-teton: Works for me, I just wasn't aware of the standard python naming for tests. `mock.py` it is.

colman-hartinger:
@bradymadams  after talking on slack earlier I decided to move both Cura and Uranium into the docker image since it made more sense in my head to keep them there instead of to be constantly cloning. If we want to test against a different version then we can implement some github action matrix and build docker images for those versions as well

bradymadams:
> @bradymadams after talking on slack earlier I decided to move both Cura and Uranium into the docker image since it made more sense in my head to keep them there instead of constantly cloning. If we want to test against a different version then we can implement some github action matrix and build docker images for those versions as well

Does the cloning inside the GH action add very much overhead? If not, I think we should consider cloning inside of the action. The reason being it will automatically test us against the latest UM and Cura without us having to keep an eye on every change in those repos. Additionally we won't have to create new docker images to keep up with the UM and Cura changes.

colman-hartinger:
> Does the cloning inside the GH action add very much overhead? 

I would say it would ~10-20 seconds but that could be offset by a quicker initialization of the image due to less layers. I'll change things to clone within the action

.github/workflows/smartslice-test.yml
```diff
@@ -0,0 +1,24 @@
+name: Smart Slice Tests
+
+on:
+  push:
+
+jobs:
+  build:
```
bradymadams, line 7: Should we rename this job to something more indicative of what it does? Maybe "unit-tests"?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Environment vars for easy configuration
+ENV CURA_APP_DIR=/srv/cura
+
+# Ensure our sources dir exists
+RUN mkdir $CURA_APP_DIR
+
+# Setup Uranium
+ENV URANIUM_BRANCH=4.6
+WORKDIR $CURA_APP_DIR
+RUN git clone -b $URANIUM_BRANCH --depth 1 https://github.com/Ultimaker/Uranium
```
bradymadams, line 12: This shouldn't be in the Dockerfile anymore, since this is done within the action, right?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Setup materials
+ENV MATERIALS_BRANCH=master
+WORKDIR $CURA_APP_DIR
```
bradymadams, line 16: I think you should only have to re-specify `WORKDIR` when it changes.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+WORKDIR $CURA_APP_DIR/Cura
+ADD . .
+RUN mv $CURA_APP_DIR/materials resources/materials
+
+# Make sure Cura can find CuraEngine
+# RUN ln -s /usr/local/bin/CuraEngine $CURA_APP_DIR/Cura docker build --tag cura_dev:1.0 .
+
+# Run Cura
+WORKDIR $CURA_APP_DIR/Cura
+ENV PYTHONPATH=${PYTHONPATH}:$CURA_APP_DIR/Uranium
```
bradymadams, line 29: This is done in the shell scripts, right?

SmartSlicePlugin/tests/test_API.py
```diff
@@ -0,0 +1,161 @@
...
+from UM.PluginRegistry import PluginRegistry
+from UM.Application import Application
+
+import pywim
+
+from SmartSliceTestCase import _SmartSliceTestCase
+
+from mock import MockJob, MockError, MockThor
+
+class test_API(_SmartSliceTestCase):
```
bradymadams, line 15: Sorry I missed this the first time around, but let's rename the class using standard class name styling.

SmartSlicePlugin/tests/run.py
```diff
@@ -0,0 +1,89 @@
...
+    def unittest_wait_and_run():
+        while wait_for_app:
+            time.sleep(0.1)
+
+        plugins = PluginRegistry.getInstance()
+
+        print("\n", file = sys.__stderr__)
+        runner = unittest.TextTestRunner(stream=sys.__stdout__, verbosity=2)
+        prog = unittest.main(exit=False, testRunner=runner, verbosity=2, argv=[sys.argv[0]])
+        print("\n", file = sys.__stderr__)
```
bradymadams, line 74: Are the two line break prints leftover from debugging?

bradymadams (changes requested):
I think the Dockerfile and shell scripts need some cleaning up. Looks like there's some redundant logic between the two. Also, use consistent styling for environment variables - it's preferred to use brackets rather than no brackets:

```bash
${PREFERRED}
$NOT_PREFERRED
```

SmartSlicePlugin/tests/test_API.py
```diff
@@ -12,7 +12,8 @@
 
 from mock import MockJob, MockError, MockThor
 
-class test_API(_SmartSliceTestCase):
+
+class testAPI(_SmartSliceTestCase):
```
bradymadams: Let's use the standard naming for classes: https://google.github.io/styleguide/pyguide.html#316-naming

Also, you guys might want to think about how you'll want to name test case classes going forward since there will ultimately be dozens of them. Choosing a good naming strategy will help you keep them organized. Something to think about, let me know if you want my opinion.

SmartSlicePlugin/tests/run.py
colman-hartinger: I tried removing the print statements and that caused a segmentation fault

SmartSlicePlugin/tests/test_API.py
bradymadams: Also, class names do not have to match the sub-module name. Again, keeping with Python standard naming conventions, module names shouldn't be camel case (testAPI), test_api would be preferred (test_API would be okay, IMO).

SmartSlicePlugin/tests/run.py
bradymadams: That seems very strange that print statements would be required to avoid a seg fault. If that's actually the case there might be a bigger problem somewhere else.

colman-hartinger: wait never mind the tests just passed without the print statements. Must've been some other error that came up when I was cleaning up some of the scripts

SmartSlicePlugin/tests/test_API.py
jake-kirkland-teton: Alright, it is done.

bradymadams:
@colman-hartinger since the Dockerfile is now just a couple of commands built on top of the cura-build-environment image - is there any reason to maintain our own image? Could those last two commands be run in the action instead?

Also, do they have a Python 3.5 image? I think it would be best if we ran our unit tests against Python 3.5 since that's what they deploy Cura with.

```
FROM ultimaker/cura-build-environment:python3.7-debian-latest

RUN apt-get install -y mesa-utils xvfb
RUN python3 -m pip install PyQt5=5.10
```

jake-kirkland-teton:
I'm actually not sure how that got included in the gitignore, that was my bad.

colman-hartinger:
> Also, do they have a Python 3.5 image? I think it would be best if we ran our unit tests against Python 3.5 since that's what they deploy Cura with.
>

@bradymadams  It doesn't look like there is a Python 3.5 image dev environment image available. Now that the Dockerfile is only 3 lines we can get rid of it and just add the packages to the build script. It will add  ~ 10 seconds to the test  time but the action is running pretty fast so I don't see a problem with that

bradymadams:
> @bradymadams It doesn't look like there is a Python 3.5 image dev environment image available. Now that the Dockerfile is only 3 lines we can get rid of it and just add the packages to the build script. It will add ~ 10 seconds to the test time but the action is running pretty fast so I don't see a problem with that

:+1: I think the added 10 seconds is worth the trade-off for not having to manage our own docker image.

bradymadams (approved):

bradymadams:
Will merge pending approval from @jake-kirkland-teton and @dougkenik

dougkenik (approved):

jake-kirkland-teton (approved):

PR #350, opened 2020-10-15
Allow plus sign in email on login

bradymadams:
This is real simple. As I was testing some logins using my gmail with a plus sign I noticed the plug-in wouldn't let me input the `+` character into the email input. This PR changes the regex validation to allow for the `+` character.


jake-kirkland-teton (approved):
Easy enough

dougkenik (approved):

PR #353, opened 2020-10-29
R20.2 custom messaging

### LINKED ISSUE #343: Custom messaging
Develop a custom messaging interface for all Smart Slice messaging.

jake-kirkland-teton:
Alright guys, this isn't perfect, but it is a foundation to improve upon and it allows us to have control over our messages. There will still need to be functionality added as we move forward, but now we have the ability to display messages in the center of the screen, with a theme font of our choosing, with the possibility of centered buttons.

I'm sure there will be some changes to make. Let me know what you guys think.

Closes #343


SmartSlicePlugin/stage/ui/SmartSliceMessage.qml
```diff
@@ -0,0 +1,515 @@
...
+                top: parent.top
+                left: parent.left
+                right: parent.right
+                margins: UM.Theme.getSize("default_margin").width
+            }
+
+            height: childrenRect.height
+
+            Button
+            {
```
dougkenik, line 102: Lots of style things in this file. In general, we always put the open bracket on the same line as the item - `Button {`

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -206,6 +209,21 @@ def onStageSelected(self):
         if self._invalid_scene_message and self._invalid_scene_message.visible:
             self._invalid_scene_message.hide()
 
+        from .ui.SmartSliceMessageExtension import SmartSliceMessage
```
dougkenik, line 212: What is this for?

dougkenik (changes requested):

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: These are only there to demonstrate the messaging system with test messages. They will be removed before the final merge.

SmartSlicePlugin/stage/ui/SmartSliceMessage.qml
jake-kirkland-teton: Fair. This is a copy of the MessageStack.qml, with additional functionality put in for our messages. I can go through it and adjust the styling.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,209 @@
...
+        if self._inactivity_timer:
+            self._inactivity_timer.stop()
+
+    def _startInactivityTimer(self) -> None:
+        if self._inactivity_timer:
+            # There is some progress indication, but no lifetime, so the inactivity timer needs to run.
+            if self._progress is not None and self._lifetime == 0:
+                self._inactivity_timer.start()
+
+    def setLifetimeTimer(self, timer: QTimer) -> None:
```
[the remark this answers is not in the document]
jake-kirkland-teton: I had a hard time getting the timers working, and this was part of my original fix to the problem. My fault, I meant to clean these up and forgot them. You are correct, we do not need a few of these.

dougkenik:
I also think there should be a way to inherit Message.qml and override what you need to. Have you searched for QML inheritance at all?

jake-kirkland-teton:
I did indeed do some research, but I may not have done a very good job. I can spend some more time looking into it.

dougkenik:
> I did indeed do some research, but I may not have done a very good job. I can spend some more time looking into it.

I think you might be able to use Message.qml in your object, and then modify properties of that - I don't know the details though:

https://stackoverflow.com/questions/11111785/does-qml-support-state-inheritance-overriding

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+
+    inactivityTimerStop = pyqtSignal()
+    inactivityTimerStart = pyqtSignal()
+    actionTriggered = Signal()
+    optionToggled = Signal()
+
+    titleChanged = Signal()
+    textChanged = Signal()
+    progressChanged = Signal()
+
```
colman-hartinger, line 59: Does using the same signal name  UM cause any problems?

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+        ALIGN_LEFT = 2
+        ALIGN_RIGHT = 3
+
+    def __init__(self, ssm_text: str = "", ssm_lifetime: int = 30, ssm_dismissable: bool = True, ssm_progress: float = None,
+                 ssm_title: Optional[str] = None, ssm_font: str = "default", ssm_use_inactivity_timer: bool = True,
+                 ssm_image_source: str = "", ssm_image_caption: str = "", ssm_option_text: str = "", ssm_option_state: bool = True) -> None:
+
+        super().__init__(text=ssm_text, lifetime=ssm_lifetime, dismissable=ssm_dismissable, progress=ssm_progress, title=ssm_title,
+                         use_inactivity_timer=ssm_use_inactivity_timer, image_source=ssm_image_source, image_caption=ssm_image_caption,
+                         option_text=ssm_option_text, option_state=ssm_option_state)
```
colman-hartinger, line 32: possibly break these functions parameters up they're a little hard to read

SmartSlicePlugin/stage/ui/SmartSliceMessage.qml
```diff
@@ -0,0 +1,515 @@
...
+                    sourceComponent:
+                    {
+                        if (modelData.button_style == 0)
+                        {
+                            return base.primaryButton
+                        } else if (modelData.button_style == 1)
+                        {
+                            return base.link
+                        } else if (modelData.button_style == 2)
+                        {
```
colman-hartinger, line 484: maybe use the SmartSliceActionButtonStyle here instead of ints. also in javascript you want to use === for comparisons although for this case it shouldn't make a difference

SmartSlicePlugin/stage/ui/SmartSliceMessage.qml
```diff
@@ -0,0 +1,515 @@
...
+                    }
+
+                    for(var index = 0; index < sizeOfActions; index++)
+                    {
+                        var actionButton = message.actions.getItem(index)
+
+                        var alignPosition = actionButton["button_align"]
+
+                        // ActionButtonStyle.BUTTON_ALIGN_CENTER == 1
+                        if (alignPosition == 1)
```
colman-hartinger, line 393: unused comment

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+        user that it is gone.
+        """
+        if self._visible:
+            self._visible = False
+            self.inactivityTimerStop.emit()
+            if send_signal:
+                self.hideSSMessageSignal.emit(str(id(self)))
+
+    def addAction(self, action_id: str, name: str, icon: str, description: str, button_style: int = SmartSliceActionButtonStyle.DEFAULT, button_align: int = SmartSliceActionButtonAlignment.ALIGN_RIGHT):
+        """Add an action to the message
```
colman-hartinger, line 83: use multi line function declaration

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+        :param button_style: Description the button style (used for Button and Link)
+        :param button_align: Define horizontal position of the action item
+        """
+        self._actions.append({"action_id": action_id, "name": name, "icon": icon, "description": description, "button_style": button_style, "button_align": button_align})
+
+    def getFont(self) -> str:
+        return self._font
+
+class SmartSliceMessageModel(VisibleMessagesModel):
+    FontRole = Qt.UserRole + 15
```
colman-hartinger, line 97: what is the + 15 for?

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+    @pyqtSlot(str)
+    def hideSSMessage(self, message_id):
+        self.removeSSMessage(message_id)
+
+    def _addRole(self, message):
+        self.appendItem({"font": message.getFont()})
+
+    def removeSSMessage(self, message_id):
+        ss_stage = SmartSliceStage.getInstance()
+        for index in range(0,len(self.items)):
```
colman-hartinger, line 154: you don't need the 0 in for the range

colman-hartinger (changes requested):
couple comments, Also all the qml should be styled according to google js style guide mainly all the semi-colons missing

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: Not that I have found.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: The role is added to the other roles from the VisibleMessagesModel that it is overriding, and they number to 14. We could change that number to one farther out to give them more space to add roles later if we choose.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: Done.

SmartSlicePlugin/stage/ui/SmartSliceMessage.qml
jake-kirkland-teton: Fair. Cleared it.

SmartSlicePlugin/stage/ui/SmartSliceMessage.qml
jake-kirkland-teton: I will look into using the ActionButtonStyle, I have some changes to attempt in the QML at this point already.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: I tried to reorganize them and added a new line

bradymadams:
The rendering isn't super clean on my machine. Anyone else see this?

image

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+        self.addRoleName(self.FontRole, "font")
+
+        SmartSliceMessage.hideSSMessageSignal.connect(self.hideSSMessage)
+
+    def _populateMessageList(self):
+        ss_stage = SmartSliceStage.getInstance()
+        for message in ss_stage.ss_messages:
+            self.addMessage(message)
+
+    def createActionsModel(self, actions):
```
bradymadams, line 110: The implementation of this function looks identical to the parent implementation. Is it necessary to override it?

bradymadams (changes requested):
I know you're working on some other changes right now, so this isn't an in depth review. Just a few surface things that I noticed. 

1. The "SS" abbreviations isn't really typical for naming in our repo. I know spelling out "SmartSlice" makes those names verbose, but it's more consistent with how we name our variables, etc.

2. The location of `ss_messages` and `visible_ss_messages` seems off to me. They are entirely managed in `SmartSliceMessageModel`, so it seems like they'd be more appropriately located within that class. Is there a reason why you need to put them in the `SmartSliceStage` class?

SmartSlicePlugin/stage/ui/SmartSliceMain.qml
```diff
@@ -1,7 +1,7 @@
 /*
     SmartSliceMain.qml
     Teton Simulation
-    Last Modified October 16, 2019
+    Last Modified October 29, 2020
```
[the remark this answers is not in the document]
jake-kirkland-teton: Deal. Removed.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+        self._lifetime_timer = None  # type: Optional[QTimer]
+        self._font = ssm_font
+
+        self._option_text = ssm_option_text
+        self._option_state = ssm_option_state
+        self._image_source = ssm_image_source
+        self._image_caption = ssm_image_caption
+
+        self._use_inactivity_timer = ssm_use_inactivity_timer
+        self._inactivity_timer = None  # type: Optional[QTimer]
```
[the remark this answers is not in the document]
jake-kirkland-teton: You are absolutely correct. I had a really hard time getting the timers and some of the functionality working at first, and I made a lot of changes and I was under the assumption that I needed all of the changes I made, but as I pare them down, a lot of these were unneeded. I have removed most of these.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,160 @@
...
+            message.progressChanged.connect(self._onMessageProgress)
+
+            ss_stage.visible_ss_messages.append(message)
+
+    @pyqtSlot(str)
+    def hideSSMessage(self, message_id):
+        self.removeSSMessage(message_id)
+
+    def _addRole(self, message):
+        self.appendItem({"font": message.getFont()})
```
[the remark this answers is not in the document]
jake-kirkland-teton: No, the signals do not actually need redefined, I have those removed.

Hmm... it does not appear that it does. I was building on the way that Cura does it, assuming it was needed at some point, but the font role still functions as expected without. I'll remove it.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: It is indeed. And no, I no longer think it is needed to override it at this point. I have removed it.

jake-kirkland-teton:
So I think overall I got a little frustrated with this and my understanding of how some of it worked. Through the requested changes and working on this a bit more, I feel like I have a bit better of a handle on it now. I appreciate you guys helping me out on this one and I apologize that it needed so much adjusting. Thanks for the patience with me as well.

@dougkenik: I have tried quite a few different methods to override the MessageStack.qml, and I have been unable to access some of the nested items that we want to adjust, such as the font/label for the messages, changing the close button action, or adding the additional action button setup for centered buttons. Do you have any ideas on how to access those objects? I have tried accessing them through encapsulation, through using dot chains, overriding the entire item structure, etc. Not having any luck on that front.

@bradymadams: On the `visible_messages` and `visible_smartslice_messages` front, I may be doing something incorrectly or not understanding properly, but the `SmartSliceMessageModel` is created AFTER the messages are attempted to be shown, which doesn't allow me to access the lists that are created in the Model, for the first messages. So I was accessing the instance of the Stage since it is created and it is a central location that I can actually get the created instance of to add and remove messages from the lists.

I worked on the formatting of the QML file overall, though it probably still needs a little work. One question on that, were we putting semicolons at the end of ALL statement lines? Or just returns and functions?

dougkenik:
> So I think overall I got a little frustrated with this and my understanding of how some of it worked. Through the requested changes and working on this a bit more, I feel like I have a bit better of a handle on it now. I appreciate you guys helping me out on this one and I apologize that it needed so much adjusting. Thanks for the patience with me as well.
> 
> @dougkenik: I have tried quite a few different methods to override the MessageStack.qml, and I have been unable to access some of the nested items that we want to adjust, such as the font/label for the messages, changing the close button action, or adding the additional action button setup for centered buttons. Do you have any ideas on how to access those objects? I have tried accessing them through encapsulation, through using dot chains, overriding the entire item structure, etc. Not having any luck on that front.
> 
> @bradymadams: On the `visible_messages` and `visible_smartslice_messages` front, I may be doing something incorrectly or not understanding properly, but the `SmartSliceMessageModel` is created AFTER the messages are attempted to be shown, which doesn't allow me to access the lists that are created in the Model, for the first messages. So I was accessing the instance of the Stage since it is created and it is a central location that I can actually get the created instance of to add and remove messages from the lists.
> 
> I worked on the formatting of the QML file overall, though it probably still needs a little work. One question on that, were we putting semicolons at the end of ALL statement lines? Or just returns and functions?

I don't know how to access the items. There should be a way to override a QML class, but UM might not have set it up that way.

bradymadams:
> @bradymadams: On the `visible_messages` and `visible_smartslice_messages` front, I may be doing something incorrectly or not understanding properly, but the `SmartSliceMessageModel` is created AFTER the messages are attempted to be shown, which doesn't allow me to access the lists that are created in the Model, for the first messages. So I was accessing the instance of the Stage since it is created and it is a central location that I can actually get the created instance of to add and remove messages from the lists.

Okay, I think I understand. And I see that your implementation is consistent with UM's implementation, which stores the messages in the Application instance. From a design perspective, it definitely breaks encapsulation, so it's not ideal. I'm not saying we should change it, it's probably fine as it is, especially since we're mocking Cura's message system.

jake-kirkland-teton:
@dougkenik Alright. The only things I have found are alias's and signals, but UM doesn't have them setup in their MessageStack. And I agree that it's a lot of code duplication, which I'm not a fan of either, but I don't see another way to get around it with the way things are currently. 

@bradymadams I agree it's not optimal. I will leave it as is and if we determine a better way to handle it down the road, we can always attempt that later.

jake-kirkland-teton:
I think this is ready for a second look over. I attempted to use a global to check against for the button alignment, but it doesn't load at the right time and causes a bizarre issue where it checks against undefined.

And @bradymadams  I do get the rendering issues at times, but it's the same issue that appears when the window is just the right size. The text rendering is already NativeRendering, so I am honestly not sure what to do to prevent that issue.

dougkenik (commented):
The text looks fine for me. I think the confirmation button needs to be blue, and the cancel / others can be grey - same styles as the validate button we have (minus the highlight).

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
bradymadams: After the first line the lines should only be indented one level and the closing parenthesis should be on it's own line aligned with the initial line:

```python
def foo(
    x: int,
    s: str
):
    # foo logic

File: SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py [hunk 15]
```diff
@@ -0,0 +1,126 @@
...
+        """Add an action to the message
+        Actions are useful for making messages that require input from the user.
+        :param action_id:
+        :param name: The displayed name of the action
+        :param icon: Source of the icon to be used
+        :param button_style: Description the button style (used for Button and Link)
+        :param button_align: Define horizontal position of the action item
+        """
+        self._actions.append({"action_id": action_id, "name": name, "icon": icon,"description": description,
+                              "button_style": button_style, "button_align": button_align})
```
bradymadams, line 72: Is there any difference between this statement and the one performed in the parent `addAction`? If not, just call the parent function.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
bradymadams: We should probably do that, if possible. It would suck if Cura released an update with an added role and that was the only thing that broke our plug-in compatibility. If possible, I'd recommend we add a large number, like 100.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,126 @@
...
+            })
+            message.titleChanged.connect(self._onMessageTitleChanged)
+            message.textChanged.connect(self._onMessageTextChanged)
+            message.progressChanged.connect(self._onMessageProgress)
+
+            smartslice_stage.visible_smartslice_messages.append(message)
+
+    @pyqtSlot(str)
+    def hideSmartSliceMessage(self, message_id):
+        self.removeSmartSliceMessage(message_id)
```
bradymadams, line 116: Is `hideSmartSliceMessage` necessary since it's just a thin proxy for `removeSmartSliceMessage`?

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,126 @@
...
+
+from ..SmartSliceStage import SmartSliceStage
+
+from UM.Logger import Logger
+
+class SmartSliceMessage(Message):
+    class SmartSliceActionButtonStyle:
+        DEFAULT = 0
+        LINK = 1
+        SECONDARY = 2
```
bradymadams, line 19: Since we're not adding anything to this enum that doesn't exist in the `Message.ActionButtonStyle` I'd say we just omit this.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,126 @@
...
+
+    class SmartSliceActionButtonAlignment:
```
bradymadams, line 21: Can you look into deriving this from `Message.ActionButtonAlignment` and just adding the `ALIGN_CENTER`. If you do that, I think you could completely omit the implementation of `addAction` in `SmartSliceMessage`.

bradymadams (changes requested):

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: Good point. I do not believe it actually is, I think our signals work a bit differently than the ones for the normal messages so we shouldn't need it. I removed it.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: Done.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: I actually think it works without overriding it. I was under the impression that it was going to be necessary but you are right, it isn't.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: I can just change this so only ALIGN_CENTER is in this class, and rename it ActionButtonAlignment, which will allow us to just call `SmartSliceMessage.ActionButtonAlignment.ALIGN_CENTER` or `Message.ActionButtonAlignment.ALIGN_LEFT/RIGHT`.

If we don't want to have UM.Message imported in these situations as well, we can always just have all three of them defined in the `SmartSliceMessage` `ActionButtonAlignment` too.

jake-kirkland-teton:
Alright, we are now using `Cura.PrimaryButton` and `Cura.SecondaryButton` for the buttons and I have made the requested changes Brady. If you feel like we should add all of the Alignment types to our class, I can do that as well, it depends on how we want to do it.

bradymadams (changes requested):
Looks good. Just one last final request. If the Doug and Colman think it looks good as well let's remove the test code and prep it for merge.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
```diff
@@ -0,0 +1,106 @@
...
+from UM.Qt.ListModel import ListModel
+from UM.Signal import Signal
+from UM.Qt.Bindings.VisibleMessagesModel import VisibleMessagesModel
+
+from ..SmartSliceStage import SmartSliceStage
+
+from UM.Logger import Logger
+
+class SmartSliceMessage(Message):
+    class ActionButtonAlignment:
```
bradymadams, line 16: Let's derive this from `Message.ActionButtonAlignment` so that it picks up the other alignment properties.

bradymadams (changes requested):
Sorry, forgot to submit the change request comment on my last review. Here it is.

SmartSlicePlugin/stage/ui/SmartSliceMessageExtension.py
jake-kirkland-teton: I am not sure why I didn't fully understand that request the first time you requested it. Sorry Brady. Got it fixed.

dougkenik (approved):
Buttons look good

packaging/package.json
```diff
@@ -9,7 +9,7 @@
     "display_name": "Smart Slice",
     "package_id": "SmartSlicePlugin",
     "package_type": "plugin",
-    "package_version": "20.2.5",
+    "package_version": "20.2.7",
```
colman-hartinger, line 12: is there a version r20.2.6?

bradymadams (approved):

colman-hartinger (changes requested):
Code looks good, sometimes after deleting a message the text rendering for the other messages gets a little funky like in Brady's photo but that doesn't seem like a real problem. The tests messages in smart slice stage should be removed before merging into r20.2

packaging/package.json
colman-hartinger: ignore that I guess the package.json version was off by one

colman-hartinger (approved):

PR #356, opened 2020-11-20
R20.2 results messages

### LINKED ISSUE #344: Result view messages

jake-kirkland-teton:
OK. The buttons are finally behaving as expected. Take a look and make sure I didn't miss anything, but this should be good to go. 

Closes #344


dougkenik (changes requested):
In general this looks really good. A few things I would like to see:

1. I don't think we need to display the error message when we cannot find an optimized solution. Instead let's choose the value which we cannot meet (fos or displacement) and enable the corresponding button / message

2. I would like to change the message for the maximum displacement both for validation and optimization. I changed the PPT to reflect this.

3. Spaces between the value and _mm_ 

4. Red text for "positions" (see PPT)

jake-kirkland-teton:
1. Alright. I can set that up. 
2. Sounds good. I will get that adjusted.
3. Easy enough.
4. Now might be a great time to mention that I am red/green colorblind, and I _thought_ I had all the colored text setup. My apologies. I will get that changed.

dougkenik:
> 1. Alright. I can set that up.
> 2. Sounds good. I will get that adjusted.
> 3. Easy enough.
> 4. Now might be a great time to mention that I am red/green colorblind, and I _thought_ I had all the colored text setup. My apologies. I will get that changed.

I see. Maybe I'll modify the colors for colorblindness later then!

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -28,6 +28,7 @@
 from UM.Math.Matrix import Matrix
 from UM.Qt.Duration import Duration
 from UM.Signal import Signal
+from UM.Message import Message
```
colman-hartinger, line 32: Shouldn't we be using the message module we built instead of Uranium's?

colman-hartinger (approved):
The messages look good and I can;t see anything I would've done differently in the code just have my one comment about using Uraniums messaging system instead of our own

jake-kirkland-teton:
Yeah, the messages are intended to be at the bottom and out of the way of the model, and not critical errors, so we are using the built in messages for this

bradymadams (commented):
I think the multiple line HTML strings are hard to read. And it makes the code hard to read when they're put in-line with the logic. I have a couple of requests:

1. Use Python's triple quote string format to create these strings.
2. Put the strings into variables and then reference them in the logic.

```python
def displayResultsMessage(self, button_string):
    safety_factor_text = """
<br>The minimum factor of safety of <font color="\{}\"><b>{}</b></font> is greater than
... and so on ..
"""
        if button_string == "stress":
            if status != SmartSliceCloudStatus.ReadyToVerify:
                if req_tool.targetSafetyFactor <= self.resultSafetyFactor:
                    text = safety_factor_text
```

I know that's not exact, but something along those lines is what I'm thinking.

And if we think this is going to be common for us to have to define large complicated HTML strings going forward I think we should look into putting this kind of content into QML. Or using HTML templates or something. I'm not sure of the best approach, but hardcoding HTML strings into Python seems like something that should be saved for "every once in awhile" situations. I think we should be careful getting complacent with it - it's going to be hard to maintain.

I'm open to thoughts and discussion on this.

jake-kirkland-teton:
OK, this has been reworked. Let me know what you guys think.

@bradymadams Doug and I talked about setting up our own messages at the bottom as well, we will probably want to do that at some point. I agree that if we start making a habit of this we should move it to a better system.

dougkenik (changes requested):
Looking much better. A few comments"

1. "Matrial" is misspelled in the text boxes.
2. Let's change "before yielding" to "before incurring permanent deformation"

jake-kirkland-teton:
Gotcha. Fixed.

bradymadams (approved):
This is definitely a cleaner look and easier to read. Nice job and thank you.

dougkenik (approved):

PR #364, opened 2020-12-10
R20.2 problem meshes

### LINKED ISSUE #345: Insert/remove problem area meshes
When applicable, add and display the problem area meshes returned by the solver for the strength and deformation views.

jake-kirkland-teton:
This adds the new problem meshes and a system to remove them so they don't interfere with other operations. 

I currently have the server set to Test since without the mesh data the new changes fail. 

Closes #345


SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -1014,6 +1004,7 @@ def processAnalysisResult(self, selectedRow=0):
 
         elif job.job_type == pywim.smartslice.job.JobType.optimization and active_extruder:
             self._proxy.resultsTable.setResults(job.getResult().analyses, selectedRow)
+            self._proxy.clearProblemMeshes()
```
dougkenik, line 1007: Why do you need to clear problem meshes here, and not after a validation? What if a user ran multiple validations?

dougkenik (changes requested):
From a UX standpoint, when a user has selected one of the results buttons (displacement or fos), leaves the stage, and then comes back, the meshes will still be present, but the message is not there, and the button is deselected. I think we should retain the selected button, meshes, and message.

Also, when viewing a problem mesh, I would suggest we also make the part more transparent, so we can see inside of it

bradymadams:
@jake-kirkland-teton any plans to add unit tests for this functionality?

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Good catch. I thought it was covered by the other clears, but I think there are a couple edge cases I didn't fully think through I need to account for.

I can certainly make the popup reappear, that shouldn't be a problem.

I will see what I can do to make the part transparent.

jake-kirkland-teton:
@bradymadams I have been thinking about getting myself into the habit of writing some unittests when I finish a ticket. I think this would be a reasonable place to start that. I will write some unittests for this system, either for this ticket, or after the displacement is added, whichever seems smarter.

bradymadams:
> @bradymadams I have been thinking about getting myself into the habit of writing some unittests when I finish a ticket. I think this would be a reasonable place to start that. I will write some unittests for this system, either for this ticket, or after the displacement is added, whichever seems smarter.

My opinion is to add the unit tests along with the functionality, in the same PR. This is what we do in our solver repos and it works really well.

colman-hartinger (approved):
All the code looks good to me. Running some tests in smartslice to test the ui

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -52,10 +52,14 @@ def beginRendering(self):
                 if node.getMeshData() and node.isVisible() and not node.callDecoration("getLayerData"):
                     per_mesh_stack = node.callDecoration("getStack")
 
-                    if node.callDecoration("isNonPrintingMesh"):
+                    if node.callDecoration("isNonPrintingMesh") and not node.hasDecoration("is_problem_area"):
                         uniforms["diffuse_color"] = [.55, .69, .1, 1]
                         uniforms["hover_face"] = -1
                         renderer.queueNode(node, shader=self._non_printing_shader, uniforms=uniforms, transparent=True)
+                    elif node.callDecoration("isNonPrintingMesh") and node.hasDecoration("is_problem_area"):
```
bradymadams, line 59: I'm not sure how efficient the node decorators are and since this logic is in the rendering loop it gets called a lot. I would pull the decorator calls outside of the if statements and store them in variables so they only get called once.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -205,10 +205,14 @@ def onStageSelected(self):
             controller.setActiveTool(use_tool)
 
         self._connector.updateSliceWidget()
+        proxy = self._connector._proxy
 
         if self._invalid_scene_message and self._invalid_scene_message.visible:
             self._invalid_scene_message.hide()
 
+        if proxy.hasProblemMeshesVisible == True:
```
bradymadams, line 213: `if proxy.hasProblemMeshesVisible:` is preferred.
https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

SmartSlicePlugin/utils/__init__.py
```diff
@@ -45,6 +50,41 @@ def makeInteractiveMesh(mesh_data: MeshData) -> 'pywim.geom.tri.Mesh':
 
     return int_mesh
 
+def buildMesh(mesh, title):
```
bradymadams, line 53: Let's add type annotations here so it's more clear what arguments it takes and what it returns.

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -900,3 +1063,18 @@ def updateSceneFromOptimizationResult(self, analysis: pywim.smartslice.result.An
...
+                if problem_region.name == mesh_to_display:
+                    problem_mesh_node = buildMesh(problem_region, "ProblemRegion")
+                    problem_mesh_node.setSelectable(False)
+                    problem_mesh_node.addDecorator(SmartSliceProblemAreaDecorator())
+                    settings_override = problem_mesh_node.getDecorator(SettingOverrideDecorator)
+                    settings_override._is_non_printing_mesh = True
+                    self.hasProblemMeshesVisible = True
+                    our_only_node.addChild(problem_mesh_node)
+                    Application.getInstance().getController().getScene().sceneChanged.emit(problem_mesh_node)
+        pass
```
bradymadams, line 1080: Looks like a leftover `pass`?

bradymadams (changes requested):
Nice job, the code looks good and clean. A few minor requests.

SmartSlicePlugin/stage/SmartSliceStage.py
jake-kirkland-teton: Touche. That is on me. Got it changed.

SmartSlicePlugin/SmartSliceCloudProxy.py
jake-kirkland-teton: Yeah, it is. I forgot to remove that.

SmartSlicePlugin/utils/__init__.py
jake-kirkland-teton: Can do.

SmartSlicePlugin/SmartSliceView.py
jake-kirkland-teton: Very good point. Got that taken care of.

jake-kirkland-teton:
Alright, so this contains many of the requested changes, and the UX should be fixed now. It is good to go for some more testing. 

There is one last change that @dougkenik and I talked about, and it is to move the building of the meshes into their own class. That way the problem meshes and the mod meshes will be something like a `SmartSliceSceneNode` and we will be able to customize them at will.
 
Once that is done, I will be setting up the unittests for this as well.

dougkenik:
The functionality is better, except for one question....

What happens if I open a validation or optimization which was saved from a previous version of the back-end?

jake-kirkland-teton:
Hmmm, I am honestly not sure. I will have to look into that.

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -35,12 +35,22 @@ def beginRendering(self):
...
+            if node.hasDecoration("is_problem_area"):
+                has_problem_area = True
+
         for node in DepthFirstIterator(scene.getRoot()):
             if isinstance(node, (BuildVolume, ConvexHullNode, Platform)):
                 continue
 
+            is_non_printing_mesh = node.callDecoration("isNonPrintingMesh")
+            has_get_layer_data = node.callDecoration("getLayerData")
+            is_problem_area = node.hasDecoration("is_problem_area")
```
bradymadams, line 52: Probably doesn't make sense to call these unless line 61 is true, right?
Two of them could go under the if statement on line 62.

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -49,20 +59,30 @@ def beginRendering(self):
...
                         pass
                     else:
                         if overlay:
                             renderer.queueNode(node, shader = self._shader, uniforms = uniforms, overlay = True)
                         else:
-                            renderer.queueNode(node, shader = self._shader, uniforms = uniforms)
+                            if has_problem_area:
+                                renderer.queueNode(node, shader = self._shader, uniforms = uniforms, transparent = True)
+                            else:
+                                renderer.queueNode(node, shader = self._shader, uniforms = uniforms)
```
bradymadams, line 85: No whitespace around `=` on argument assignments. I know you were just using the style that was already in this file, but the general rule is when modifying existing code let's improve the style, if necessary.

https://google.github.io/styleguide/pyguide.html#36-whitespace

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -49,20 +59,30 @@ def beginRendering(self):
...
                         uniforms["diffuse_color"] = [.55, .69, .1, 1]
                         uniforms["hover_face"] = -1
-                        renderer.queueNode(node, shader=self._non_printing_shader, uniforms=uniforms, transparent=True)
-                    elif per_mesh_stack and per_mesh_stack.getProperty("support_mesh", "value"):
+                        renderer.queueNode(node, shader = self._non_printing_shader, uniforms = uniforms, transparent = True)
+                    elif is_non_printing_mesh and is_problem_area:
+                        uniforms["diffuse_color"] = [.945, .373, .388, 1]
+                        uniforms["hover_face"] = -1
+                        renderer.queueNode(node, shader = self._non_printing_shader, uniforms = uniforms, transparent = True)
+                        has_problem_area = True
```
bradymadams, line 75: I don't think this is necessary since it's set in the for loop on line 42.

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -49,20 +59,30 @@ def beginRendering(self):
...
+                    elif per_mesh_stack and is_support:
                         pass
                     else:
                         if overlay:
                             renderer.queueNode(node, shader = self._shader, uniforms = uniforms, overlay = True)
                         else:
-                            renderer.queueNode(node, shader = self._shader, uniforms = uniforms)
+                            if has_problem_area:
+                                renderer.queueNode(node, shader = self._shader, uniforms = uniforms, transparent = True)
```
bradymadams, line 83: I think this could be condensed to one call to `queueNode` using `transparent=has_problem_area`.

bradymadams (changes requested):
I think the view class could use a little more tidying up.

On second look are we adding/removing the problem meshes rather than building them once and then just showing/hiding them? Is there a reason the current approach is preferred?

SmartSlicePlugin/SmartSliceView.py
jake-kirkland-teton: Hmm. Good point. I moved is non printing and is problem area.

SmartSlicePlugin/SmartSliceView.py
jake-kirkland-teton: Ah, yes. Good catch. I meant to remove that one.

SmartSlicePlugin/SmartSliceView.py
jake-kirkland-teton: Ohhhh. I like that.

SmartSlicePlugin/SmartSliceView.py
jake-kirkland-teton: Fixed

jake-kirkland-teton:
One last comment, is that we no longer need the `SmartSliceDecorators` with this change. I did not yet remove the file, but if we are all ok with that I can remove it before we merge these changes.

bradymadams:
> One last comment, is that we no longer need the `SmartSliceDecorators` with this change. I did not yet remove the file, but if we are all ok with that I can remove it before we merge these changes.

I am always in favor of removing unused/unnecessary code. It will always be in the git history if we ever need to resurrect it.

dougkenik (commented):
So those decorators did a lot for tracking added / removed meshes and identifying Z-Offsets. Have you tested all of that?

jake-kirkland-teton:
@dougkenik I believe I have. There is a replacement added/removed system now and the Z-Offsets are a Cura Decorator that is still implemented. It would be good to get someone to double check and ensure I haven't unknowingly broken something, but I think everything is still tracked properly.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -679,3 +686,140 @@ def run(self):
...
+        self._is_removed = False
+
+        if self._mesh_type == self.MeshType.ProblemMesh:
+            self._buildProblemMesh(mesh, title)
+
+        if self._mesh_type == self.MeshType.ModifierMesh:
+            self._buildModifierMesh(mesh, title)
+
+    @property
+    def mesh_type(self):
```
dougkenik, line 708: It might be easier just to expose 'self._mesh_type' as 'self.mesh_type' and forgo the properties?

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -679,3 +686,140 @@ def run(self):
...
+        return self._mesh_type
+
+    @mesh_type.setter
+    def mesh_type(self, value: MeshType):
+        if self._mesh_type != value:
+            self._mesh_type = value
+
+    @property
+    def is_removed(self):
```
dougkenik, line 717: Same here. The properties don't make a whole bunch of sense to me

dougkenik (changes requested):
This has come a long way Jake!

dougkenik (changes requested):
Also, when I try to open the attached 3MF, I get a crash (traceback below):

2020-12-23 09:00:48,679 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [66]: An uncaught error has occurred!
2020-12-23 09:00:48,684 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]: Traceback (most recent call last):
2020-12-23 09:00:48,687 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\cura\CuraApplication.py", line 1013, in event
2020-12-23 09:00:48,691 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\UM\Qt\QtApplication.py", line 464, in event
2020-12-23 09:00:48,694 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\UM\Event.py", line 218, in call
2020-12-23 09:00:48,698 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\UM\Signal.py", line 332, in __performEmit
2020-12-23 09:00:48,701 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\stage\SmartSliceScene.py", line 516, in _process_mesh_analysis
2020-12-23 09:00:48,704 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:     job.callback()
2020-12-23 09:00:48,707 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\SmartSliceExtension.py", line 300, in afterSmartSliceNodeInit
2020-12-23 09:00:48,710 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:     self.proxy.problem_area_results = results.fea_results.steps[0].increments[0].model_results.problem_regions
2020-12-23 09:00:48,712 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\3rd-party\cpython-common\pywim\__init__.py", line 99, in __getitem__
2020-12-23 09:00:48,715 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:     return super().__getitem__(key)
2020-12-23 09:00:48,718 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]: IndexError: list index out of range
2020-12-23 09:00:48,855 - DEBUG - [MainThread] cura.Machines.Models.QualityProfilesDropDownMenuModel._update [63]: Updating QualityProfilesDropDownMenuModel.
2020-12-23 09:00:48,862 - DEBUG - [MainThread] cura.Machines.Models.CustomQualityProfilesDropDownMenuModel._update [34]: Updating CustomQualityProfilesDropDownMenuModel.
2020-12-23 09:00:50,627 - INFO - [MainThread] cura.Machines.MachineErrorChecker._setResult [187]: Error check finished, result = False, time = 3.1s
2020-12-23 09:00:56,609 - DEBUG - [MainThread] cura.AutoSave._onTimeout [60]: Autosaving preferences, instances and profiles
[webinar_validation.zip](https://github.com/tetonsim/is-cura-ui/files/5736071/webinar_validation.zip)

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -786,14 +812,52 @@ def updateStatusFromResults(self, job: pywim.smartslice.job.Job, results: pywim.
...
+
+    def showProblemMeshes(self):
+        problem_meshes = getProblemMeshes()
+        if self.message_type == "stress":
+            meshes_to_show = ("low_safety_factor")
+        else:
+            meshes_to_show = ("high_strain")
+        for mesh in problem_meshes:
+            if mesh.getName() in meshes_to_show:
+                mesh.setVisible = True
```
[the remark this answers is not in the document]
jake-kirkland-teton: Ah, you are absolutely correct.

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -178,9 +180,43 @@ def onSavePromptClicked(self):
             self.onEscapeSavePromptClicked()
 
     def _writeState(self, output_object=None):
+        self._outputManager = Application.getInstance().getOutputDeviceManager()
+        self._outputManager._output_devices["local_file"].writeSuccess.connect(self.readd_meshes)
```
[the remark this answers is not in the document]
jake-kirkland-teton: Yup. That is my bad on the naming front. Got it fixed.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Sure. I will certainly do that instead. I did a little research into it since I honestly wasn't super sure on when to and when not to use properties and you are correct, they don't really offer much benefit in this circumstance.

jake-kirkland-teton:
Alright, this should be fixed now! I also moved the SmartSliceMessages per @dougkenik 's request.

bradymadams (changes requested):
The unit tests seem pretty lacking to me. It seems like they're just creating objects and then checking a few attributes on them. I'd like to see some tests that get more code coverage on the new code.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -679,3 +686,122 @@ def run(self):
         time.sleep(1)
         self.interactive_mesh = makeInteractiveMesh(self.mesh_data)
 
+class SmartSliceMeshNode(SceneNode):
+    class MeshType(Enum):
+        Normal = 0
+        ModifierMesh = 1
+        ProblemMesh = 2
+
+    def __init__(self, mesh_type: MeshType = MeshType.Normal, mesh: 'pywim.chop.mesh.Mesh' = pywim.chop.mesh.Mesh(), title: str = "NormalMesh"):
```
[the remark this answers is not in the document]
jake-kirkland-teton: Alright. That sounds good. I would like a run down on why, just so I know, but no rush.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -679,3 +686,122 @@ def run(self):
...
+
+        modifier_mesh_node_infill_pattern = SmartSliceJobHandler.INFILL_SMARTSLICE_CURA[modifier_mesh.print_config.infill.pattern]
+        definition_dict = {
+            "infill_mesh" : True,
+            "infill_pattern" : modifier_mesh_node_infill_pattern,
+            "infill_sparse_density": modifier_mesh.print_config.infill.density,
+            "wall_line_count": modifier_mesh.print_config.walls,
+            "top_layers": modifier_mesh.print_config.top_layers,
+            "bottom_layers": modifier_mesh.print_config.bottom_layers,
+            }
```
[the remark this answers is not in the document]
jake-kirkland-teton: Gotcha. Fixed.

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        pass
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
```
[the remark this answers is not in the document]
jake-kirkland-teton: I have removed them.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -679,3 +686,122 @@ def run(self):
...
+        active_build_plate = Application.getInstance().getMultiBuildPlateModel().activeBuildPlate
+        self.addDecorator(BuildPlateDecorator(active_build_plate))
+
+        bottom = self.getBoundingBox().bottom
+
+        z_offset_decorator = ZOffsetDecorator()
+        z_offset_decorator.setZOffset(bottom)
+        self.addDecorator(z_offset_decorator)
+
+        return self
```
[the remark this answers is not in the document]
jake-kirkland-teton: I am not totally sure what I was thinking. I will make that into it's own variable that is returned rather than self.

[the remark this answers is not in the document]
bradymadams: If you do that, that implies this function should be creating a new SmartSliceMeshNode. Are you sure that's what you want?

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -679,3 +686,122 @@ def run(self):
...
+
+    def _buildProblemMesh(self, problem_mesh_data: 'pywim.chop.mesh.Mesh', mesh_to_display: str):
+        our_only_node =  getPrintableNodes()[0]
+
+        if problem_mesh_data is not None:
```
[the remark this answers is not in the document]
jake-kirkland-teton: No it does not. It was a hold over from a different way I was building the meshes. I have removed it.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -233,6 +249,12 @@ def onStageDeselected(self):
         for mesh in getModifierMeshes():
             mesh.setSelectable(True)
 
+        proxy = self._connector._proxy
+        if proxy.results_buttons_popup_visible:
+            proxy.closeResultsButtonPopup()
+            proxy.results_buttons_popup_visible = True
```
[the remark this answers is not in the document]
jake-kirkland-teton: It is necessary. `closeResultsButtonPopup` does in fact set it to False. I thought it would be more efficient to just reuse the proxy variable so the popup can automatically come back up when you swap back to the tab, rather than writing a new method to handle it.

jake-kirkland-teton:
Alright, so I have made a fair amount of changes. There are still a couple things to be done when I return:

1) Figure out the parent for the `SmartSliceMeshNode` class. I tried a couple things but was unsucessful.
2) Test the changes to the unittests and then build some new more effective ones and work with Brady to make that happen.

jake-kirkland-teton:
OK. After some more research and testing, I have found that there is no need to pass in a constructed scene node as the parent. The ONLY reason a parent is passed into the UM SceneNode constructor is to do an automatic `.addChild` to that parent on construction, and our nodes will always be added to the parent of our choosing after creation anyway.
Additionally, I was unable to get the `SmartSliceSceneNode` constructor to work correctly directly passing the node we want as the parent in, because of some signal emitting issues on inheriting the other `init`s, so I believe the way it is being constructed and attached to it's desired parent with this last commit is the correct way to handle it, unless I have missed something.

bradymadams:
> Additionally, I was unable to get the `SmartSliceSceneNode` constructor to work correctly directly passing the node we want as the parent in, because of some signal emitting issues on inheriting the other `init`s, so I believe the way it is being constructed and attached to it's desired parent with this last commit is the correct way to handle it, unless I have missed something.

So that means the problem meshes don't have a parent, right? It feels more appropriate if their parent was the smart slice scene root or the printable node they are related to.

jake-kirkland-teton:
Actually it just means they aren't assigned a parent within the constructor for the object. When they are added to the scene they are added as a child to the `only_printable_node`, which then becomes their parent.

bradymadams:
@jake-kirkland-teton are you ready for another round of reviews on this or are you still working on unit tests?

jake-kirkland-teton:
I am indeed working on better unittests. Though unless that is absolutely critical to the release, I would suggest that we continue with the reviews and keep getting the features in so it can get to a point where the team can start testing it. I will keep working on tests and bugfixes and the displacement stuff as we move forward.

bradymadams:
Is your suggestion that we move forward to merge this without better unit tests?

jake-kirkland-teton:
If that is acceptable, yes. If the consensus is to wait until I complete better tests, we can do that as well. I just don't want to be holding the testing of the release up.

bradymadams:
> If that is acceptable, yes. If the consensus is to wait until I complete better tests, we can do that as well. I just don't want to be holding the testing of the release up.

I'd prefer we add good unit tests along with the functionality in each PR. We've been dragging our feet for too long on writing more unit tests in the plug-in - we need to start making it a standard part of the development process.

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -786,14 +812,52 @@ def updateStatusFromResults(self, job: pywim.smartslice.job.Job, results: pywim.
...
+            self.deflectionOpacity = 0.5
+            self.results_buttons_popup_visible = False
+            self.resetResultsButtons()
+
+    def showProblemMeshes(self):
+        problem_meshes = getProblemMeshes()
+        if self.message_type == "stress":
+            meshes_to_show = ("low_safety_factor")
+        else:
+            meshes_to_show = ("high_strain")
```
bradymadams, line 834: This is just setting a string, but it looks like you actually want a tuple. One item tuples need a comma.

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -786,14 +812,52 @@ def updateStatusFromResults(self, job: pywim.smartslice.job.Job, results: pywim.
...
+        for mesh in getProblemMeshes():
+            mesh.setVisible(False)
+            Application.getInstance().getController().getScene().sceneChanged.emit(mesh)
+
+    @pyqtSlot()
+    def removeProblemMeshes(self):
+        problem_meshes = getProblemMeshes()
+        if problem_meshes != []:
+            our_only_node =  getPrintableNodes()[0]
+            if len(problem_meshes) > 0:
```
bradymadams, line 851: Is this redundant with the check on line 849?

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -679,3 +686,125 @@ def run(self):
...
+            if value is not None:
+                definition = stack.getSettingDefinition(key)
+                property_type = type_map.get(stack.getProperty(key, "type"))
+                if property_type:
+                    new_instance = SettingInstance(definition, settings)
+
+                    new_instance.setProperty("value", property_type(value))
+
+                    new_instance.resetState()  # Ensure that the state is not seen as a user state.
+                    settings.addInstance(new_instance)
```
bradymadams, line 805: Looks like the above functionality is very similar to that in SmartSliceCloudProxy.py: https://github.com/tetonsim/is-cura-ui/pull/364/files#diff-bcd48d4cb072bf3a670395eadeb84f6f25ffa67ea5ba92f37d99ba158d0a5313L942

This logic is complicated enough that my first reaction is that it should be consolidated. Thoughts?

bradymadams (changes requested):
Looking better. A few items I noticed. I'll review again once there's updates to the unit tests.

SmartSlicePlugin/SmartSliceCloudProxy.py
jake-kirkland-teton: Ah yes, you are correct. Fixed.

SmartSlicePlugin/SmartSliceCloudProxy.py
jake-kirkland-teton: I think I was just being extra cautious. I don't think we need the second check. Removed.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Now that I am comfortable with the tests, I will look into this a little more. I think it might be able to be consolidated. They are very similar, I just need to try to move them somewhere accessible and try to generalize it so it can be used in both cases.

SmartSlicePlugin/SmartSliceCloudProxy.py
bradymadams: The check is still there in the PR. Maybe you have a commit you haven't pushed yet?

SmartSlicePlugin/tests/mock.py
```diff
@@ -107,4 +108,31 @@ def get_token(self):
...
+        cura_instance = CuraApplication.getInstance()
+        machine_manager = cura_instance.getMachineManager()
+        machine_manager.addMachine("ultimaker_s5")
+
+        root = cura_instance.getController().getScene().getRoot()
+
+        self._node = SceneNode(root, name="Test_cube")
+
+        cube = MeshBuilder()
+        cube.addCube(1,1,1)
```
bradymadams, line 130: spaces between arguments

SmartSlicePlugin/tests/mock.py
```diff
@@ -107,4 +108,31 @@ def get_token(self):
...
+        cube_mesh = cube.build()
+        self._node.setMeshData(cube_mesh)
+        self._node.addDecorator(SliceableObjectDecorator())
+        #with patch("cura.")
+        self._node.addDecorator(SettingOverrideDecorator())
+
+        pass
```
bradymadams, line 137: "pass" not necessary here

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -0,0 +1,233 @@
...
+        mesh_data[0].name = "high_strain"
+
+        self.testing_proxy.problem_area_results = mesh_data
+        self.testing_proxy.resultMaximalDisplacement = 2.0
+        self.testing_proxy.resultSafetyFactor = 2.0
+        self.testing_proxy.deflectionOpacity = 1.0
+        self.testing_proxy.displayResultsMessage("deflection")
+
+        our_mesh = getProblemMeshes()
+        self.assertNotEqual(our_mesh, [])
```
bradymadams, line 97: If we expect a certain number of meshes in the returned list then I think it would be better to assert the length of the list.

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -0,0 +1,233 @@
...
+    @patch("SmartSlicePlugin.stage.SmartSliceScene.SmartSliceMeshNode.getBoundingBox", MagicMock())
+    def test_smartslice_high_strain_mesh(self):
+        from SmartSlicePlugin.utils import getProblemMeshes
+
+        mesh = pywim.chop.mesh.Mesh(name="high_strain")
+        mesh.vertices = [threemf.mesh.Vertex(x=1, y=2, z=3), threemf.mesh.Vertex(x=4, y=5, z=6), threemf.mesh.Vertex(x=7, y=8, z=9)]
+        mesh.triangles = [threemf.mesh.Triangle(v1=0, v2=1, v3=2)]
+        mesh_data = [pywim.fea.result.ModelRegion()]
+        mesh_data[0].regions = [mesh]
+        mesh_data[0].name = "high_strain"
```
bradymadams, line 88: The above few lines are duplicated a few times in this class - I'd recommend consolidating into a simple static method.

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -0,0 +1,233 @@
...
+import pywim
+import threemf
+
+from SmartSliceTestCase import _SmartSliceTestCase
+
+class TestSmartSliceMesh(_SmartSliceTestCase):
+    @classmethod
+    def setUpClass(cls):
+        from SmartSlicePlugin.SmartSliceCloudProxy import SmartSliceCloudProxy
+        new_node = MockModel()
```
bradymadams, line 21: It doesn't look like this is used anywhere. If not, is the `MockModel` class needed?

bradymadams (changes requested):
The unit tests look much better.

SmartSlicePlugin/tests/mock.py
jake-kirkland-teton: Yeah, it's on my current commit that I haven't pushed yet.

Gotcha, I meant to go back and clean up the formatting a little and forgot.

SmartSlicePlugin/tests/mock.py
jake-kirkland-teton: Another place I didn't clean up. I was using the pass to debug.

SmartSlicePlugin/tests/test_smartslice_mesh.py
jake-kirkland-teton: Ah fair enough. I think that is a good point, that makes it easy to modify once the deflection meshes are in too.

SmartSlicePlugin/tests/test_smartslice_mesh.py
jake-kirkland-teton: Touche. I will do that.

SmartSlicePlugin/tests/test_smartslice_mesh.py
jake-kirkland-teton: The model is not technically used directly, but the `MockModel` class builds a model for the build plate to emulate a loaded stl. I suppose I could just call MockModel rather than assign it to a variable if that would be cleaner.

jake-kirkland-teton: And actually, the method doesn't return anything, so I should just be calling it anyway.

bradymadams: Well MockModel is a class, not a method, so it does return an object of `MockModel`. But if the only use of it is the side effects that occur from the logic in the constructor then it might be cleaner to just move it to a free function, or a static method in the test class.

jake-kirkland-teton: Alright. That is a good point. I was thinking of keeping it available to other future tests, but that is a good point, I think we could add it to the `_SmartSliceTestCase` class as a method.

bradymadams: This doesn't need to be done in this PR, but something you might think about is how to design a system that will let you easily and consistently setup the state of Cura for unit tests. You can imagine as unit tests are added managing the state (nodes, print settings, Smart Slice state, etc.) will get tricky. Offloading that into classes/functions that are dedicated to cleaning up the state from a previous unit test and setting it back up will be very helpful.

jake-kirkland-teton:
Alright. I have the required changes in. I think this is ok for another look over.

SmartSlicePlugin/tests/mock.py
```diff
@@ -3,6 +3,7 @@
 from typing import Callable
 
 import pywim
+import threemf
```
bradymadams, line 6: Unused?

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -0,0 +1,226 @@
...
+        self.testing_proxy.resultMaximalDisplacement = 0.0
+        self.testing_proxy.results_buttons_popup = None
+        self.testing_proxy.results_buttons_popup_visible = False
+        self.testing_proxy.previous_message = ""
+        self.testing_proxy.deflectionOpacity = 0.5
+        self.testing_proxy.stressOpacity = 0.5
+        self.testing_proxy.result_feasibility = None
+        self.testing_proxy._sliceStatusEnum = SmartSliceCloudStatus.Underdimensioned
+
+    def makeProblemMeshData(self, name_string):
```
bradymadams, line 35: If `self` isn't used in a class function then making it `static` is preferred.

We usually don't include type information in variable/argument names. Using a type annotation is preferred.

SmartSlicePlugin/utils/__init__.py
```diff
@@ -141,3 +153,24 @@ def intersectingNodes(node: CuraSceneNode) -> List[CuraSceneNode]:
                 intersecting_nodes.append(n)
 
     return intersecting_nodes
+
+def updateProperties(property_dict: dict, func: callable, property_object, property_object_type: str):
```
bradymadams, line 157: [callable is an actual Python built-in function](https://docs.python.org/3.5/library/functions.html#callable). Use the `Callable` type from the typing module and don't forget to include the argument types with it.

I like type annotations, so I like when I see them added to the code, but I think they're most helpful when the whole function is annotated and not just partially. So I'd add one for `property_object` as well.

And using `dict` is fine for a type annotation if any generic dictionary is acceptable, but if there's any required structure to it, using `Dict` from `typing` would be preferred. [See this for more info](https://stackoverflow.com/questions/37087457/difference-between-defining-typing-dict-and-dict).

bradymadams (changes requested):
I think the new location of the update properties logic looks fine. Maybe the name of the function could use some massaging so it's a little more clear what it's doing. What's your concern with it?

SmartSlicePlugin/tests/mock.py
jake-kirkland-teton: Ah yeah it is. Good catch.

SmartSlicePlugin/tests/test_smartslice_mesh.py
jake-kirkland-teton: Oh I see. I was not aware. That makes sense. I will get it switched over.

I will remember that. Do you require that I go in and change the code in proxy to fix this? I am only setting an existing variable.

bradymadams: Not sure I know where in the code you're referring to, or which of my comments you're referring to.

But generally, style updates to code that is not changed in the current PR are a judgment call. I think it's fine to hit a few spots for cleanup, but making dozens of changes can sometimes overwhelm an already complex PR.

If you're referring to the static method change request - in Python that is somewhere between a software design and style issue. I'd be more apt to change that if I came across it.

https://tetoncomposites.atlassian.net/wiki/spaces/DEV/pages/357040133/Coding+Style

jake-kirkland-teton: Sorry that was unclear. I got the static method changed. That one is on me.

The one in question was the naming of `_sliceStatusEnum`, which is used in the Proxy class.

SmartSlicePlugin/utils/__init__.py
jake-kirkland-teton: Fair enough. I was not clear on the `callable/Callable` distinction. 

I thought about the `property_object` annotation, but it requires a couple of different types of objects depending on where the function is called from. How do I handle the that in that situation? In one case it is an `ExtruderStack`, in another case it is a `PerObjectContainerStack`. Do I just call it a generic `Stack`?

And alright. I will read that post and it sounds likely that it will be a `Dict` from `typing` instead. Thanks!

SmartSlicePlugin/tests/test_smartslice_mesh.py
bradymadams: That looks like it touches way too much code for it to be a harmless style update.

Every code base is going to have imperfections that don't align with our adopted style or design philosophies. Usually it's from legacy code (created before an adopted style, etc) or was missed in a PR review. This is one of those cases. Sometimes it might be harmless to update it, other times it might be risky. Since this particular one affects QML and many other Python files, I think it's much too risky.

jake-kirkland-teton: I agree with that assessment and that is what I was thinking. That is why I wanted to clarify if you were expecting I change it for the PR. I certainly would have done it if asked.

SmartSlicePlugin/utils/__init__.py
bradymadams: For `property_object` you should ask yourself what types will work and what won't work (will cause an exception). If anything that derives from `Stack` will work with the function then that's the proper type. If not, and it _must_ be one of the other specific two stack types then you should use [Union](https://docs.python.org/3/library/typing.html#typing.Union).

And now that we've clarified what the type likely is, this points out another place where naming could be improved. If the type is `Stack` then it would be a lot more clear if the argument was named `stack` or something similar. Names like property, object, key, etc. are ambiguous. Sometimes they're unavoidable when writing truly abstract code, but they should be avoided if at all possible.

Think about coming across your code as a developer and trying to figure out what it does. If you see:

```python
def foo1(stack: Stack):

def foo2(property_object):
```

`foo1` is _way_ more clear on how I use it.

jake-kirkland-teton: I see. Alright. That actually makes sense. I believe in this case the function will accept any `CuraContainerStack` so I will be labeling it as such. 

And that is a very good point. The names are ambiguous. I have changed them to be clearer, once I commit them let me know what you think of the new names. I agree, your example shows a very good point. Thank you for taking the time to help me learn.

SmartSlicePlugin/utils/__init__.py
```diff
@@ -141,3 +155,24 @@ def intersectingNodes(node: CuraSceneNode) -> List[CuraSceneNode]:
...
+        "str": str,
+        "enum": str,
+        "bool": bool
+    }
+
+    definition = None
+
+    for key, value in property_dict.items():
+        if value is not None:
+            if container_stack_type == "settings_stack":
```
bradymadams, line 172: I know this logic was already there, but sometimes as code gets cleaned up other things become more apparent. Generally, it's not great design to use strings to control logic flow. Again, as a user of this function it's not immediately obvious to me what I should be sending for `container_stack_type` - I would have to read the code to understand what it's used for. Compare to something like:

```python
def updateContainerStackProperties(...., get_setting_definition: bool):
```

This would be a little more clear to me. However, I'm questioning why the call to `getSettingDefinition` even needs to be conditional. Couldn't it just be called every time regardless of the "stack type"?

bradymadams (commented):
The types look good.

SmartSlicePlugin/utils/__init__.py
```diff
@@ -141,3 +155,24 @@ def intersectingNodes(node: CuraSceneNode) -> List[CuraSceneNode]:
                 intersecting_nodes.append(n)
 
     return intersecting_nodes
+
+def updateContainerStackProperties(property_dict: Dict[str, Any], func: Callable[[CuraContainerStack, str, Any, type, SettingDefinition], None], container_stack: CuraContainerStack, container_stack_type: str):
```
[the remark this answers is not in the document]
jake-kirkland-teton: Will do.

SmartSlicePlugin/utils/__init__.py
jake-kirkland-teton: Hmmm. Fair point. I think we absolutely could call it every time regardless. I think I was assuming it was specific to the `PerObjectContainerStack` but even then I think just checking to ensure it isn't `None` would be fine. I'll adjust that and test it.

bradymadams (approved):

bradymadams:
Will merge pending approval from @dougkenik

dougkenik (changes requested):
Some user experience things I am noticing:

1. What happens if I open a result from a previous saved validation / optimization where the "problem meshes" were not returned? It looks like we are trying to display something still? My suggestion would be to not show the results buttons if we open a project without the appropriate data (e.g. from a previous release)
2. If I run a job with the new code, and then open a job from a previous release without problem meshes, the problem meshes returned from the first job are still present. This makes me think we are not clearing the data when we open a project?

jake-kirkland-teton:
1. I was thinking we could still show the popup with the extra instructions and information in that case, which is why the buttons are left there. I completely understand how that could be a bit of a confusing/poor UX so I will remove the buttons.
2. Hmmm. I had thought I was but that might not be the case. I will investigate.

dougkenik (changes requested):
Run an optimization that fails to optimize, save it, and then open it. Cura crashes with the below message:


Traceback (most recent call last):
  File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\cura\CuraApplication.py", line 1013, in event
  File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\UM\Qt\QtApplication.py", line 464, in event
  File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\UM\Event.py", line 218, in call
  File "X:\4.6-exe\build\inst\lib\python3.5\site-packages\UM\Signal.py", line 332, in __performEmit
  File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\stage\SmartSliceScene.py", line 516, in _process_mesh_analysis
    job.callback()
  File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\SmartSliceExtension.py", line 270, in afterSmartSliceNodeInit
    self.cloud.processAnalysisResult(selected_row)
  File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\SmartSliceCloudConnector.py", line 1005, in processAnalysisResult
    resultData = ResultTableData.analysisToResultDict(0, job.getResult().analyses[0])
  File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\3rd-party\cpython-common\pywim\__init__.py", line 99, in __getitem__
    return super().__getitem__(key)
IndexError: list index out of range

jake-kirkland-teton:
Yeah I discovered this one a bit ago. It happens without the problem meshes too. I started a ticket. 
https://app.zenhub.com/workspaces/smartslice-5ef3719d317b88001a6dc21e/issues/tetonsim/is-cura-ui/368

dougkenik (approved):
Ok. The UI looks good now.

PR #366, opened 2020-12-14
R20.2 tooltip

dougkenik:
Added all of the tooltips, except the description on Anchor / Load buttons (this is much harder, and I don't see the ROI right now). The work includes:

- Built out a Tooltip which is more flexible than the standard Cura tooltip
- Added the tooltip to the appropriate dialogs and the functions to show / hide
- Built out a function for reporting metadata in the Results Table (used in the tooltip)


jake-kirkland-teton (approved):
This looks really good. I think they look good, and I think they function very well. The code is clean and well done.

SmartSlicePlugin/components/SmartSliceTooltip.qml
```diff
@@ -0,0 +1,112 @@
...
+            x = point.x + UM.Theme.getSize("default_arrow").width;
+            y = point.y - 0.5 * UM.Theme.getSize("default_arrow").height - UM.Theme.getSize("tooltip_arrow_margins").height;
+        }
+
+        base.opacity = 1;
+    }
+
+    function hide() {
+        base.opacity = 0;
+    }
```
colman-hartinger, line 53: Is there a better way to hide the tooltip than setting the opacity to 0?

dougkenik: I copied what Cura does. Using opacity allows us to define a behavior on opacity changing, which results in the tooltips fading in and out

SmartSlicePlugin/stage/ui/ResultTable.py
```diff
@@ -178,6 +178,34 @@ def data(self, index, role):
...
+    def getResultMetaData(self, row) -> str:
+        result = self._results[row]
+        metadata = """<p> Extruder: </p>
+            <p style = 'margin-left:50px;'> Walls: {0:d} </p>
+            <p style = 'margin-left:50px;'> Top Layers: {1:d} </p>
+            <p style = 'margin-left:50px;'> Bottom layers: {2:d} </p>
+            <p style = 'margin-left:50px;'> Infill density: {3:.1f}% </p>""".format(
+                int(round(result.print_config.walls, 0)),
+                int(round(result.print_config.top_layers, 0)),
+                int(round(result.print_config.bottom_layers, 0)),
```
bradymadams, line 191: Shouldn't these already be ints? Why do we have to round and cast them?

bradymadams (commented):
Looks good, but some redundancy in the re-implementation of `showTooltip` and `hideTooltip` functions. And also in the `onEntered` and `onExited` event handlers. I'm not a QML guru, so I don't know what's all possible, but it would be nice to see some abstraction and consolidation of these.

SmartSlicePlugin/stage/ui/ResultTable.py
dougkenik: They are not ints... For some reason they are floats (coming from Smart Slice). I had to cast them to ints to stop the program from crashing

dougkenik:
> Looks good, but some redundancy in the re-implementation of `showTooltip` and `hideTooltip` functions. And also in the `onEntered` and `onExited` event handlers. I'm not a QML guru, so I don't know what's all possible, but it would be nice to see some abstraction and consolidation of these.

Yes, it's a pain in the butt. The only other way I see to do it is to have the Stage handle all of the displaying of the tooltip, but that makes getting the correct x & y locations very difficult. Basically, the way it is implemented now allows us to use the QML Item's local CS to display the tooltip. Putting it in the stage would mean we would have to map the CS to the stage CS, which doesn't always work as designed.

SmartSlicePlugin/stage/ui/ResultTable.py
bradymadams: We should probably create an issue to investigate why that's happening.

bradymadams:
> Yes, it's a pain in the butt. The only other way I see to do it is to have the Stage handle all of the displaying of the tooltip, but that makes getting the correct x & y locations very difficult. Basically, the way it is implemented now allows us to use the QML Item's local CS to display the tooltip. Putting it in the stage would mean we would have to map the CS to the stage CS, which doesn't always work as designed.

I guess what I was thinking was creating a Tooltip component that extends MouseArea (or something similar). It could overlay a textbox and capture the entered and exited events. It would contain properties that define the text and the location, so all we would need to do is add that component and set the properties where necessary.

dougkenik:
> > Yes, it's a pain in the butt. The only other way I see to do it is to have the Stage handle all of the displaying of the tooltip, but that makes getting the correct x & y locations very difficult. Basically, the way it is implemented now allows us to use the QML Item's local CS to display the tooltip. Putting it in the stage would mean we would have to map the CS to the stage CS, which doesn't always work as designed.
> 
> I guess what I was thinking was creating a Tooltip component that extends MouseArea (or something similar). It could overlay a textbox and capture the entered and exited events. It would contain properties that define the text and the location, so all we would need to do is add that component and set the properties where necessary.

I don't think that's possible, because the component of which we want the tooltip to originate from needs to be extended to have the mouse area. This is why I had to create the `HoveredTextField `component, and why UM created the `UM.SimpleButton` component - these components extend traditional components and implement `MouseArea` which then shows / hides a Tooltip - but the Tooltip component needs to be shown (most of the time) on the Item/Component which holds the component that enacted the hover events (if that makes sense). I could be missing a very logical architectural change though

bradymadams:
I think what I'm asking for is actually mostly there, but I think we should offload more of the functionality to the Tooltip component via properties. Which probably means having more than one tooltip component. So defining the location, text, etc. as properties, rather than defining the entered/exited event on each component that needs a tooltip and then overriding the properties of the one and only tooltip.

dougkenik:
Ok, so you basically would want a tooltip for every component which would enact one? So, every button, text field, etc. would have an associated (and individualized) tooltip tied to it instead of using one tooltip and moving it around / setting the text?

bradymadams:
> Ok, so you basically would want a tooltip for every component which would enact one? So, every button, text field, etc. would have an associated (and individualized) tooltip tied to it instead of using one tooltip and moving it around / setting the text?

Yes. Maybe there are other implications with that, but on the surface it seems a lot more maintainable to simply define a tooltip via properties, rather than redefining events and manipulating the same tooltip over and over, which will have the potential for bugs to sneak in that depend on the order of events.

dougkenik:
I'll dive into it. I'm afraid it will cause more confusion than what it is worth due to the various Coordinate System things get defined in, but I cannot say.

dougkenik:
Ok - so I have updated the implementation as follows:

- Added a tooltip to the SmartSlice.HoverableTextField component
- Created a new component called SmartSlice.HoverableButton which is an extension of UM.SimpleButton with a tooltip
- Modified all the text fields and buttons with their own tooltip
- For the table, I simplified it a bit, but I kept one tooltip for the entire table instead of one tooltip per row. IMO, this is easier to code, read, and change in the future.

bradymadams (approved):

colman-hartinger (approved):
code looks and and looks good in cura

jake-kirkland-teton (approved):
Looks slick and functions well.

bradymadams (approved):

PR #371, opened 2021-01-15
R20.2 display deformed mesh

### LINKED ISSUE #346: Display deformed mesh
Use the displacements of the surface mesh vertices returned by the solver to display a deformed mesh.

jake-kirkland-teton:
Closes #346


dougkenik (changes requested):
Deformed mesh should be displayed even if the displacement meets requirements.

jake-kirkland-teton:
Gotcha. Should be fixed now.

dougkenik (commented):
My only concern here is that it's hard to see the problem mesh when the displacement mesh is inserted. Is there an order to how these are displayed? If so, maybe we display the problem meshes after the displacement mesh?

image

dougkenik:
It's still hard to see the problem area mesh when the displacement is on. Any ideas on how to make it a bit clearer?

We might explore putting a way for the user to turn on or off the displacement mesh or problem mesh when viewing the displacement... thoughts?

jake-kirkland-teton:
Hmmm. Did my change make a difference for you? It made a world of difference for me. I just want to ensure we are seeing the same things. Here are some screenshots:
Old - 
image

New - 
image

On ideas on making it clearer, I'm not quite sure the best course of action. Coloring might be an option, maybe different colors? I'm not sure if we can change opacity, there might some limitations we would need to figure out. The toggle could also work, we might need another button however.

dougkenik:
You are correct. I flubbed the download / install of the new package I built. I am OK with where it is now, but let's get feedback from the team to see if we need something else.

As far as the toggle, I was thinking we could put a button in the popup to show/hide the displaced mesh. Thoughts?

dougkenik:
The other question I had is that it doesn't seem like you're "normalizing" the displacement when they come back. My fear is that a user will apply a very low load, and won't be able to see the displacement because it will basically lie on top of the first mesh... or, conversely, they apply a ridiculously large load and the displacement is massive? I believe @bradymadams has implemented some logic to work around some of these before

jake-kirkland-teton:
For the toggle, I think that could probably be accomplished. Though those messages are already pretty big. There will probably have to be some additional work done to split how the meshes are being handled to show or hide one or the other, but it is doable.

And on normalizing, no I am not doing any at the moment. We can certainly look into it. Once the displacements get really large they start doing strange things anyway, and when they are small, they do indeed look like they haven't moved much, though the color does change. I'll reach out to @bradymadams.

dougkenik:
> For the toggle, I think that could probably be accomplished. Though those messages are already pretty big. There will probably have to be some additional work done to split how the meshes are being handled to show or hide one or the other, but it is doable.
> 
> And on normalizing, no I am not doing any at the moment. We can certainly look into it. Once the displacements get really large they start doing strange things anyway, and when they are small, they do indeed look like they haven't moved much, though the color does change. I'll reach out to @bradymadams.

If we need another ticket for the toggle based on team feedback, we can create one later. The normalized displacement should be implemented for this ticket though.

jake-kirkland-teton:
That sounds good. I talked to Brady about the normalization, he is going to help me write it a bit later when he has some time.

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -990,7 +991,8 @@ def _onJobFinished(self, job):
                         self._proxy.resultsButtonsVisible = True
                         self._proxy.removeProblemMeshes()
                         self._proxy.problem_area_results = self._jobs[self._current_job].getResult().fea_results.steps[0].increments[0].model_results.problem_regions
-                        self._proxy.result_feasibility = self._jobs[self._current_job].getResult().feasibility_result["structural"]
+                        self._proxy.displacement_mesh_results = self._jobs[self._current_job].getResult().surface_mesh_results.steps[0].increments[0].node_results
+                        self._proxy.result_feasibility = self._jobs[self._current_job].getResult().feasibility_result['structural']
```
bradymadams, line 995: I think there's some opportunity for cleanup in this function. There's a lot of repeated calls to the same functions and attributes. Mainly `self._jobs[self._current_job].getResult()`.

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -989,19 +990,25 @@ def displayResultsMessage(self, button_string):
                 if status != SmartSliceCloudStatus.ReadyToVerify:
                     if req_tool.maxDisplacement >= self.resultMaximalDisplacement:
                         text = displacement_acceptable
+                        SmartSliceMeshNode(self.displacement_mesh_results, SmartSliceMeshNode.MeshType.DisplacementMesh, "displacement_mesh")
+                        self.hasProblemMeshesVisible = True
```
bradymadams, line 994: Is this attribute always being set to `True` now? If so, can it be pulled out of all these nested if statements?

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -256,12 +258,14 @@ def afterSmartSliceNodeInit():
             if self.cloud.status == SmartSliceCloudStatus.Optimized:
                 self.cloud.addJob(pywim.smartslice.job.JobType.optimization)
                 if results.fea_results is not None and results.fea_results.steps != []:
-                    self.proxy.problem_area_results = results.fea_results.steps[0].increments[0].model_results.problem_regions
+                    self.proxy.problem_area_results = results.fea_results.steps[0].increments[0].model_results.problem_region
+                    self.proxy.displacement_mesh_results = results.surface_mesh_results.steps[0].increments[0].node_resultss
```
bradymadams, line 262: Looks like some misspelled attributes here.

SmartSlicePlugin/SmartSliceView.py
```diff
@@ -62,9 +62,12 @@ def beginRendering(self):
             if not node.render(renderer):
                 is_non_printing_mesh = node.callDecoration("isNonPrintingMesh")
                 is_problem_area = False
+                is_displacement_mesh = False
                 if isinstance(node, SmartSliceMeshNode):
                     if node.mesh_type == SmartSliceMeshNode.MeshType.ProblemMesh:
                         is_problem_area = True
+                    if node.mesh_type == SmartSliceMeshNode.MeshType.DisplacementMesh:
```
bradymadams, line 69: I think `elif` would be more appropriate here.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -32,7 +33,7 @@
 from cura.Scene.SliceableObjectDecorator import SliceableObjectDecorator
 from cura.Settings.SettingOverrideDecorator import SettingOverrideDecorator
 
-import pywim
+import pywim, threemf
```
bradymadams, line 36: [We usually put imported modules on separate lines](https://google.github.io/styleguide/pyguide.html#313-imports-formatting).

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -704,10 +706,13 @@ def __init__(self, mesh: "pywim.chop.mesh.Mesh", mesh_type: MeshType = MeshType.
...
             self._buildModifierMesh(mesh, title)
 
+        if self.mesh_type == self.MeshType.DisplacementMesh:
+            self._buildDisplacementMesh(mesh)
+
         if self.mesh_type == self.MeshType.Normal:
             self._buildMesh(mesh, title)
 
-    def _buildMesh(self, mesh: "pywim.chop.mesh.Mesh", title: str) -> "SmartSliceMeshNode":
+    def _buildMesh(self, mesh: 'pywim.chop.mesh.Mesh', title: str) -> 'SmartSliceMeshNode':
```
bradymadams, line 715: Double quotes.

Also, if the type is imported there's no reason to use quotes, and I think some IDEs have an easier time finding the type if it's not in quotes. Quotes on types are sometimes necessary, but I don't think it is for the `Mesh` type here.

There's other places in this file that should probably be checked for this.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -796,3 +803,39 @@ def _updateSettings(self, stack, key, value, property_type, definition):
...
+        model_mesh = getPrintableNodes()[0]
+
+        mesh_vertices = []
+        mesh_triangles = []
+        mesh_data = pywim.chop.mesh.Mesh()
+        for displacement in displacement_mesh[0].values:
+            mesh_vertices.append(threemf.mesh.Vertex(
+                x=displacement.data[0] + model_mesh._mesh_data._vertices[displacement.id][0],
+                y=displacement.data[2] + model_mesh._mesh_data._vertices[displacement.id][1],
+                z=-displacement.data[1] + model_mesh._mesh_data._vertices[displacement.id][2]
```
bradymadams, line 816: Anytime you're accessing a private attribute you should question why you're doing it and if there's a better way. It should really be reserved for only exceptional cases. In this case, I'm pretty sure there are functions for both of these attributes, so those would be the preferred way to access the data.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -796,3 +803,39 @@ def _updateSettings(self, stack, key, value, property_type, definition):
...
+                    v1=ind.data[0],
+                    v2=ind.data[1],
+                    v3=ind.data[2]
+                ))
+        else:
+            mesh_triangles = None
+
+        mesh_data.vertices = mesh_vertices
+        mesh_data.triangles = mesh_triangles
+        mesh_data.transform = [[1, 0, 0, 0],[0, 1, 0, 0],[0, 0, 1, 0],[0, 0, 0, 1]]
```
bradymadams, line 831: Spaces between arguments.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -796,3 +803,39 @@ def _updateSettings(self, stack, key, value, property_type, definition):
...
+
+        self._buildMesh(mesh_data, "displacement_mesh")
+
+        settings_override = SettingOverrideDecorator()
+        settings_override._is_non_printing_mesh = True
```
bradymadams, line 836: Do we do this anywhere else in our code base? I had a look at `SettingOverrideDecorator` and changing this attribute directly seems questionable on the surface. It looks like this is usually evaluated by checking some particular settings.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -796,3 +803,39 @@ def _updateSettings(self, stack, key, value, property_type, definition):
...
+        self.addDecorator(settings_override)
+
+        our_only_node = getPrintableNodes()[0]
+        our_only_node.addChild(self)
+        Application.getInstance().getController().getScene().sceneChanged.emit(self)
```
bradymadams, line 841: Out of curiosity - why do we need to emit this signal? Looks like `addChild` emits some more specific signals.

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -43,6 +44,17 @@ def makeProblemMeshData(name_string):
...
+    @staticmethod
+    def makeDisplacementResultData(proxy):
+        displacement_result = [pywim.fea.result.Result(name="displacement")]
+        vals = []
+        for i in range(0, 8):
+            val = pywim.fea.result.ResultValue(data=[0.1, 0.1, 0.1], id=i)
+            vals.append(val)
+        displacement_result[0].values = vals
+
+        proxy.displacement_mesh_results = displacement_result
```
bradymadams, line 56: A general comment on function design. Consider the following two functions:

```python
def foo(obj: SomeClass):
  a = ... code to build up some complex variable ...
  obj.a = a

def bar() -> A:
  a = ... code to build up some complex variable ...
  
c1 = SomeClass()
foo(c1)

c2 = SomeClass()
c2.a = bar()

a = bar()
```

`bar` is more flexible and can be used in more cases and doesn't restrict it from being used in the same way `foo` is.

This isn't critical, but it was succinct enough of a function that I just wanted to take the opportunity to point out.

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -43,6 +44,17 @@ def makeProblemMeshData(name_string):
 
         return mesh_data
 
+    @staticmethod
+    def makeDisplacementResultData(proxy):
+        displacement_result = [pywim.fea.result.Result(name="displacement")]
+        vals = []
+        for i in range(0, 8):
```
bradymadams, line 51: `range(N)` is preferred over `range(0, N)`.

SmartSlicePlugin/tests/test_smartslice_mesh.py
```diff
@@ -43,6 +44,17 @@ def makeProblemMeshData(name_string):
...
+            val = pywim.fea.result.ResultValue(data=[0.1, 0.1, 0.1], id=i)
+            vals.append(val)
+        displacement_result[0].values = vals
```
bradymadams, line 54: A general code style comment: sometimes using temporary variables makes sense, but I wouldn't always default to it. It can help code readability to omit the temporaries if they aren't replacing something complicated and can reduce the number of lines.

bradymadams (changes requested):

SmartSlicePlugin/SmartSliceCloudProxy.py
jake-kirkland-teton: Ah yeah it is. Good point. I moved it. Is it acceptable to set it after the displaying of the meshes? I prefer to always set things like that after actually doing what the attribute is representing, in case something fails along the way.

SmartSlicePlugin/SmartSliceExtension.py
jake-kirkland-teton: You are correct. Something happened during one of the merges and I lost a few style changes and fixing this spelling error. I remember this one specifically. I'll go back and fix it and the quotes.

SmartSlicePlugin/SmartSliceView.py
jake-kirkland-teton: Can do.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Understood. I'll fix that.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: You are absolutely correct. I need to be better about that. I'm not sure why it didn't occur to me to fix that one alongside the `getIndices` change. I'll get it fixed.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Fixed.

SmartSlicePlugin/tests/test_smartslice_mesh.py
jake-kirkland-teton: Ohh right. I forgot it could be done that way. Changed.

colman-hartinger:
@jake-kirkland-teton Are there any changes I need to make to my config to see the deformed mesh? This is all I got when I just tested things out
displacement

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Alright. I cleaned this up.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Hmmm. I do not think we do. 

I can adjust this to set one of the settings related to ensure the mesh ends up a non printing mesh, however it appears our options are to set it to "anti_overhang_mesh", "infill_mesh" or "cutting_mesh". I'm not sure what I would want to set it to there. What do you guys think?

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: I am honestly not positive we do need to emit this signal. It was a part of the modifier mesh displaying/removing system and I just kept it when I moved it. `addChild` definitely sends a few signals. I can look into the signals sent to ensure they are covering the same things.

SmartSlicePlugin/tests/test_smartslice_mesh.py
jake-kirkland-teton: Fair enough. Thank you for the info. That makes sense. I have changed it in this case, since it was indeed simple. But I do see why that is much more useful.

SmartSlicePlugin/tests/test_smartslice_mesh.py
jake-kirkland-teton: Understood. I had not thought about that. I will do my best to keep that in mind going forward. I know I usually default to using a named variable. I'll work on that one. I changed it in this instance, and one more.

SmartSlicePlugin/SmartSliceCloudConnector.py
jake-kirkland-teton: Alright. I cleaned this up and moved some of the duplicated code into it's own function.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Upon researching, I do not believe we need the emit. Unless someone can find a reason otherwise, everything seems to function the same without it. I am fairly sure `addChild` calls `sceneChanged` down the line anyway. I have removed them.

jake-kirkland-teton:
Alright @bradymadams, all of the changes requested are done, with the exception of the one regarding `SettingsOverrideDecorator`. I believe the best course of action will be to consider our displacement/problem meshes as an `"anti_overhang_mesh"` simply because it seems to exclude the most excess code we do not care about executing. If not, we can also consider them `"infill_mesh"`es. I would like thoughts on the subject. (@bradymadams, @dougkenik)

SmartSlicePlugin/stage/SmartSliceScene.py
dougkenik: I think using "anti_overhang_mesh" or "infill_mesh" seems questionable. Is there another way to make a mesh a "non printing mesh"? I thought we did that with our Smart Slice Node?

bradymadams: @jake-kirkland-teton in SmartSliceView.py would `is_displacement_mesh` ever be True when `is_non_printing_mesh` is False? It seems like the check for `is_non_printing_mesh` isn't even necessary, but maybe I'm missing something.

jake-kirkland-teton: After looking at it a little closer, I believe Brady is correct. I do not think we actually need to set the mesh as non-printing. The reasons being are:
1. The meshes are always removed before any other operations are performed in regards to slicing/printing.
2. The only real thing we want is to render them as non-printing, which can be achieved in the way Brady described. 

I have changed it to reflect this. Check it out and if you see any problems I might have overlooked let me know, but I think this is ok the way it is now.

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -245,6 +245,8 @@ def _getState(self, filename=None):
 
         self.cloud.clearJobs()
 
+        check = getPrintableNodes()
```
colman-hartinger, line 248: is this check ever used?

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -1,8 +1,9 @@
-from typing import List, Any, Union
+from typing import List, Any, Union, Dict
 from enum import Enum
```
colman-hartinger, line 2: Dict doesn't seem to ever be used along with a couple other imports in this file

colman-hartinger (commented):
Code looks good and everything runs as expected and looks good just a few small requests.

@dougkenik I think the UX is a little cluttered with the red problem areas showing in the displacement tab and I also think it could be good to change the color to make it slightly different from the infill mesh color

bradymadams (commented):
Current code looks good. Let's address the items Colman pointed out and I'll review again after the displacement normalization logic is in place.

SmartSlicePlugin/SmartSliceExtension.py
jake-kirkland-teton: It is not, leftover debugging. Good catch.

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Ah yeah, a lot of code has moved around and I hadn't cleaned the imports up yet. They are fixed now.

jake-kirkland-teton:
Changes made. Normalization in. The normalization cut-offs are easily changed, but I feel they are in a decent place after testing a few parts. Let me know if you feel differently.

dougkenik (changes requested):
I built the package and installed. When I open Cura the Smart Slice tab doesn't appear. The log is below:

2021-01-20 11:52:13,691 - ERROR - [MainThread] UM.Logger.logException [110]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\__init__.py", line 24, in <module>
2021-01-20 11:52:13,694 - ERROR - [MainThread] UM.Logger.logException [110]:     from . import SmartSliceExtension, SmartSliceView
2021-01-20 11:52:13,696 - ERROR - [MainThread] UM.Logger.logException [110]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\SmartSliceExtension.py", line 20, in <module>
2021-01-20 11:52:13,698 - ERROR - [MainThread] UM.Logger.logException [110]:     from .SmartSliceCloudConnector import SmartSliceCloudConnector
2021-01-20 11:52:13,701 - ERROR - [MainThread] UM.Logger.logException [110]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\SmartSliceCloudConnector.py", line 31, in <module>
2021-01-20 11:52:13,703 - ERROR - [MainThread] UM.Logger.logException [110]:     from .SmartSliceCloudProxy import SmartSliceCloudProxy
2021-01-20 11:52:13,705 - ERROR - [MainThread] UM.Logger.logException [110]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\SmartSliceCloudProxy.py", line 20, in <module>
2021-01-20 11:52:13,707 - ERROR - [MainThread] UM.Logger.logException [110]:     from .SmartSliceProperty import SmartSlicePropertyColor
2021-01-20 11:52:13,710 - ERROR - [MainThread] UM.Logger.logException [110]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\SmartSliceProperty.py", line 14, in <module>
2021-01-20 11:52:13,712 - ERROR - [MainThread] UM.Logger.logException [110]:     from .stage.SmartSliceScene import SmartSliceMeshNode
2021-01-20 11:52:13,714 - ERROR - [MainThread] UM.Logger.logException [110]:   File "C:\Users\DougKenik\AppData\Roaming\cura\4.6\plugins\SmartSlicePlugin\stage\SmartSliceScene.py", line 6, in <module>
2021-01-20 11:52:13,717 - ERROR - [MainThread] UM.Logger.logException [110]:     from statistics import median
2021-01-20 11:52:13,719 - ERROR - [MainThread] UM.Logger.logException [110]: ImportError: No module named 'statistics'

jake-kirkland-teton:
Alright Doug, that one is fixed. Should be good to go for testing.

dougkenik (changes requested):
Did you remove the import that was causing the crash? I still cannot get the plugin to work when I build it and then install in Cura

jake-kirkland-teton:
I'm an idiot. One second

dougkenik (approved):
This looks good now from a XD standpoint

bradymadams (changes requested):
The logic for normalizing the displacements looks good to me. I have two requests:

1. Arbitrary constants should usually go into a variable. If they are specific to a class (which I would say they are in this case) then I like to put them under the class. [And recall that they are named in uppercase](https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations). I'm referring to the `0.25` and `1.1` in this case.
2. There is one potential place for a crash. It should be unlikely, but the division of the `max_displacement` should be checked for zero. This shouldn't happen if the user is creating a good model and the solver is behaving, but shit happens. So you should check if `max_displacement` is less than or equal to `sys.float_info.min`, and if it is just set `normalize_scalar` to 1.

jake-kirkland-teton:
Should be good now Brady. Good points.

bradymadams (approved):

bradymadams:
Will merge pending approval from @colman-hartinger

colman-hartinger (approved):
everything looks good to me

PR #373, opened 2021-01-15
skin_angles & infill_angles list parsing crash bugfix

AijunHall:
`skin_angles = SettingFunction(skin_angles)(self._propertyHandler)` is the problem line from `UM/UM/Settings/SettingFunction.py`

If the given `str` from `self._propertyHandler.getExtruderProperty("skin_angles")` doesn't include the correctly formatted brackets `[]`, or if an element in the 'to-be-list' has a preceeding 0 as shown in the ticket gif, then 
`SettingFunction(skin_angles)(self._propertyHandler)` will return `NoneType` instead of a `list`.

With verifying the `list` type beforehand here, Cura won't crash. And 'invalid' inputs to the skin angles will be brought up by `UM/UM/Settings/SettingFunction.py` 

Here is the error brought up in the debugger, (no crash):
```
ERROR - [MainThread] UM.Settings.SettingFunction.__init__ [67]: Parse error in function (INPUTTED_SKIN_ANGLES) for setting: invalid token (eval, line 1)
```


AijunHall:
One thing to consider, this doesn't account for when a user would enter an integer into the skin_angles. It would go to the default `else: print_config.skin_orientations.extend((45, 135))`

I can change that, but I just went off the setting description: 'A list of integer line directions'

image

jake-kirkland-teton:
I believe it might be a better strategy to have our code reset the showing value to a valid one in the event of an invalid input, rather than just sending the default values back to us. That way, the user doesn't think they have entered a value that is not actually representative of what we are using to calculate. Thoughts @bradymadams?

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -168,7 +168,7 @@ def checkJob(self, machine_name="printer", show_extruder_warnings=False) -> Tupl
         skin_angles = self._propertyHandler.getExtruderProperty("skin_angles")
         if type(skin_angles) is str:
             skin_angles = SettingFunction(skin_angles)(self._propertyHandler)
-        if len(skin_angles) > 0:
+        if type(skin_angles) is list and len(skin_angles) > 0:
```
bradymadams, line 171: I realize the code above is doing the same thing, but it must be pretty old, because we prefer using [isinstance](https://docs.python.org/3/library/functions.html#isinstance) now.

However, I think `if skin_angles:` would be False in both the case that skin_angles is None and when it's a list with 0 length.

bradymadams (changes requested):
This is a tough one, as @jake-kirkland-teton has alluded to. I don't think we want to set the skin orientations to a value that the user might not be expecting, and I don't think we can reset the property because we could be resetting it while the user is editing it which would be annoying UX.

When the skin_angles property can't be parsed and comes back as None I would consider keeping the current logic you have, but also add a message to the returned `errors`. This means one of two things would happen:

1. If the user is in the middle of editing the skin_angles, and ultimately change it to something valid the error will eventually go away.
2. If the user has edited skin_angles to something invalid we will be setting an error indicating such and they'll be notified that they need to fix it before they can proceed.

AijunHall:
> I would consider keeping the current logic you have, but also add a message to the returned errors.

I just borrowed the same error handling code that I saw throughout this script for this skin_angles field @bradymadams .  I still do the `list` type check, since in the event that the user enters an `int` value, it raises a fatal error when it attempts to add it to the `skin_orientations` for not being iterable.

> That way, the user doesn't think they have entered a value that is not actually representative of what we are using to calculate.

Now with this setup error, the user should only be using their own valid inputs since they'll be made aware when invalid inputs are raised @jake-kirkland-teton

jake-kirkland-teton (changes requested):
This change is good, I like the use of disabling the validate button and returning the feedback to the user that way. Just a last couple changes and I think this will be good to go:

1. While I was testing this, I changed the other "list" property and got the same crash (`"Infill Line Directions"`). We should catch this crash as well. I think that is the only other `list` property, but we should double check.

2. In the error message we should have an example of a `list` for non-programmers so it is clearer what we are expecting the user to put in. Something helping specify the brackets, like "(e.g. [90])" should work, but I think someone who is not clear what a `list` is might struggle to fix the problem.

bradymadams (changes requested):
@jake-kirkland-teton good catch. @AijunHall let's make the same fix for the infill line directions in this PR.

As for the message let's just go with: "Enter a valid list of angles". The line directions are only shown in the Expert settings, so anyone mucking around in those should be able to figure it out, or at least reset the property and start over.

bradymadams:
@AijunHall I also forgot to tell you about version incrementing. In each PR you'll want to increment the revision number of the version. Our versions are `MAJOR.MINOR.REVISION`, so you want to increment the third number. In this repo you'll find the version number in two files (both need updated):

```
packaging/package.json
SmartSlicePlugin/plugin.json
```

The current version is 20.2.12 so you'll change it to 20.2.13 in this PR.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -190,13 +190,19 @@ def checkJob(self, machine_name="printer", show_extruder_warnings=False) -> Tupl
...
+            else:
+                if len(infill_angles) > 1:
+                    Logger.log("w", "More than one infill angle is set! Only the first will be taken!")
+                    Logger.log("d", "Ignoring the angles: {}".format(infill_angles[1:]))
+                print_config.infill.orientation = infill_angles[0]
         else:
-            if len(infill_angles) > 1:
-                Logger.log("w", "More than one infill angle is set! Only the first will be taken!")
-                Logger.log("d", "Ignoring the angles: {}".format(infill_angles[1:]))
-            print_config.infill.orientation = infill_angles[0]
```
bradymadams, line 193: This shouldn't be removed, it is an important check separate from the invalid list issue.

bradymadams (changes requested):
Please include the property name in the error message like you had before: `"Invalid <i>Top/Bottom Line Directions</i>: {}".format(skin_angles)`

It's important the user know which property is creating the error.

SmartSlicePlugin/SmartSliceJobHandler.py
AijunHall: Did I remove these lines? I thought I just moved 190-193 to be within the new `list` type check, where its now lines 197-200.

Because the crash is caused by the `len()` statements, so shouldn't these lines be within the `list` check?

bradymadams: You're right, not sure how I missed that. Sorry about that.

AijunHall:
for `infill_angles` I changed the error message slightly since this field only has 1 list value. The existing error message that tells the user to only input 1 list value for the `infill_angles` comes up as well.

image

bradymadams (approved):
Looks good to me. Will merge pending Jake's approval.

jake-kirkland-teton (commented):
I like it. Looks good to me! Nice work.

jake-kirkland-teton (approved):

Issue #375, opened 2021-01-18
Job Fails on Open


dougkenik:
The attached job (3MF) was ran by a customer and saved. When they opened it, it caused a hard crash. It looks like the projects was saved after a failed run, but I might be wrong.

Let's also add a unit test to catch this scenario.

Reach out to Doug for the file

AijunHall:
> It looks like the projects was saved after a failed run

I've found a way to reproduce this crash, which is along the lines of this, and is probably what the customer did in this situation.

1. Open a new import/project without any previous work done on it so that it has no SmartSlice data
2. Set the anchor, load, and other settings so that it is ready to be validated in the SmartSlice stage 
`SmartSlicePlugin.SmartSliceCloudConnector.status` should be set to `SmartSliceCloudStatus.ReadyToVerify` *
3. Save the project without validating
4. Opening the project 3mf file results in the crash

Heres my Traceback:
```
Traceback (most recent call last):
  File "C:\dev\Cura\cura\CuraApplication.py", line 1063, in event
    return super().event(event)
  File "C:\Users\AijunHall\source\repos\UM\UM\Qt\QtApplication.py", line 464, in event
    event._function_event.call()
  File "C:\Users\AijunHall\source\repos\UM\UM\Event.py", line 218, in call
    self._function(*self._args, **self._kwargs)
  File "C:\Users\AijunHall\source\repos\UM\UM\Signal.py", line 332, in __performEmit
    func(dest, *args, **kwargs)
  File "C:\dev\Cura\plugins\SmartSlicePlugin\stage\SmartSliceScene.py", line 516, in _process_mesh_analysis
    job.callback()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSliceExtension.py", line 263, in afterSmartSliceNodeInit
    if results.fea_results is not None and results.fea_results.steps != []:
AttributeError: 'NoneType' object has no attribute 'fea_results'
```

bradymadams:
@AijunHall can you test this and check if this was fixed in your PR?

AijunHall:
@bradymadams @dougkenik This file still has a problem that I forgot to mention, thanks for reopening this. It opens without crashing now, but it fails to validate with this error:
```
SmartSlicePlugin.SmartSliceCloudConnector.submitSmartSliceJob [531]: An error occured while sending and receiving cloud job: Error while processing the job:<br>Analysis unexpectedly failed.
```
So now this Issue is that it fails on verification.

[...]

PR #379, opened 2021-01-21
no optimize results load crash bugfix

AijunHall:
This fix is simple, and prevents a crash while also saving the user's anchors, loads, and settings, but it might not be targeting the root of the problem.

The crash is caused because in this scenario, where a user opens a project they saved at the point of being told there are no optimized results, `job.getResult().analyses` returns an empty list and can't be indexed.

The reason I say it might not be targeting the root of the problem, this might have something to do with the activity of `SmartSliceCloudConnector.status` . Maybe this is why the analyses is lost because of a 'bad' activity status update?

This "no optimize results load scenario" is the only point at which I see this type of status change.
image

Not too sure since I'm still prowling over the repo, but I want to bring it to attention.


bradymadams (commented):
Would the `job.job_type` ever be `optimization` in a similar scenario? If so, I think a similar check would be needed on line 1011, or in the `setResults` function.

I also wonder if we need to change the save functionality so that incomplete results are never saved anyway, thus making it so this conditional evaluates to False:

https://github.com/tetonsim/is-cura-ui/blob/27461705a03a6cbe9256ed02b73132fa6f85237d/SmartSlicePlugin/SmartSliceExtension.py#L267-L270

jake-kirkland-teton:
@bradymadams I am doing a couple tests to try to get that situation, but I think if the results are saved with the table the `.analyses` will always be present, though it may make sense to add the check there, because, as you previously stated, shit happens.

On the subject of rewriting the saving system, I think that is up to @dougkenik  to make that call. If we want to properly save the results of a "Cannot Optimize" scenario so we can load them again, there will need to be some changes made to ensure the data is saved, perhaps a new status added. If not, I think your adjustment would work well.

dougkenik:
I don't think we should save the results of a failed job. We should revert back to the setup of that job and save it

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -1002,14 +1002,19 @@ def setResultsFromJob(self, job_result):
 
     def processAnalysisResult(self, selectedRow=0):
         job = self._jobs[self._current_job]
+        analyses = job.getResult().analyses
+        if not analyses:
+            Logger.log("d", "No analyses to process")
+            return
```
AijunHall, line 1008: Would this be a better catch to include the optimization `job.job_type`?

AijunHall:
> Would the job.job_type ever be optimization in a similar scenario?

@bradymadams In my limited messing around, unless optimization succeeds with
`SmartSliceCloudStatus.BusyOptimizing -> SmartSliceCloudStatus.Optimized`

the `job.job_type` won't be `optimization`

When the optimization fails the status gets kicked back
`SmartSliceCloudStatus.BusyOptimizing -> SmartSliceCloudStatus.ReadyToVerify`

and there isn't a job type update.  Let's see what comes of @jake-kirkland-teton 's testing

> I also wonder if we need to change the save functionality so that incomplete results are never saved anyway, thus making it so this conditional evaluates to False:

I'm also glad you brought this up, because this is the problem area for #375 where `results` is an unexpected `None` when saving a project before having a complete `JobType.validation`

_This line comes before your snippet with the `if results` and causes the crash in 375_

https://github.com/tetonsim/is-cura-ui/blob/1a503283749ca6d1105d972503c6c22d737000bc/SmartSlicePlugin/SmartSliceExtension.py#L258

So the relation between these two issues is probably the save functionality because of these `result` attributes being lost when the user does a weird combination of inputs then saves.

jake-kirkland-teton:
So Aijun is correct, I was unable to get a job_type to be `optimized` and not have `.analyses`. That doesn't mean it can't happen, but I tried a few different weird things and never caused a crash.

On the subject of saving incomplete results, since Doug mentioned we should only save the Boundary Conditions and not the results of we fail to optimize, I feel we should go with Brady's suggestion. 
@AijunHall  Do you think you could write a reasonable check to ensure the results are present or prevent saving them to resolve both issues?

AijunHall:
> Do you think you could write a reasonable check to ensure the results are present or prevent saving them to resolve both issues?

@jake-kirkland-teton Sure thing.

bradymadams (commented):
I'm not sure I understand the change in SmartSliceExtension.py. Can you explain it? It seems like if the conditional on line 256 is False then line 270 is going to be problematic (will `cloudJob` be None?).

AijunHall:
@bradymadams I see the problem now.

The previous commit does fix the 375 crash, but `cloudJob` does come out as None in the "no validation load" scenario

The crash is caused by a `results.fea_results` check when `result` is None, so I thought nesting the lines within the
256 conditional would be a reasonable fix. 

I'll simplify it. I'm also looking into the save functionality, which might be a better approach.

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -208,18 +208,20 @@ def _saveState(self, writing_workspace=False):
...
+            if cloudJob and cloudJob.getResult():
+                self._storage.setEntryToStore(plugin_id=self.metadata.id, key="results", data=cloudJob.getResult().to_dict())
+                self._storage.setEntryToStore(
+                    plugin_id=self.metadata.id,
+                    key="selectedResult",
+                    data=self.proxy.resultsTable.getSelectedResultId()
+                )
+                if writing_workspace:
+                    cloudJob.saved = True
+            elif job.type == pywim.smartslice.job.JobType.validation and (not cloudJob or not cloudJob.getResult()):
```
bradymadams, line 222: Why is it limited to `validation` here?

AijunHall: No idea to be honest. I assumed `optimization` would be caught in the first conditional. I'll add it in.

AijunHall:
> I think if we use the ReadyToValidate to prevent the results from saving in that case, it should work.

Would using the status of the result be better than using the cloud status @jake-kirkland-teton ? For the optimize issue, this is the scenario where the `no_solution_found` status would be brought up specifically, vs the broad application of `ReadyToValidate`
(to prevent problematic results from downloading)

https://github.com/tetonsim/pywim/blob/be6eddab85d08c414f1f544b0584a87fcdc9bc80/pywim/smartslice/result.py#L8-L11

bradymadams:
This looks good to me, but I just requested Doug's review on this. He understands the saving logic better than I do.

jake-kirkland-teton:
> 
> 
> > I think if we use the ReadyToValidate to prevent the results from saving in that case, it should work.
> 
> Would using the status of the result be better than using the cloud status @jake-kirkland-teton ? For the optimize issue, this is the scenario where the `no_solution_found` status would be brought up specifically, vs the broad application of `ReadyToValidate`
> (to prevent problematic results from downloading)
> 
> https://github.com/tetonsim/pywim/blob/be6eddab85d08c414f1f544b0584a87fcdc9bc80/pywim/smartslice/result.py#L8-L11

I think that would work well actually. I would double check to ensure that status is always present when you need to check for it, but as long as it is, I think that is a great idea.

AijunHall:
Good point to bring up, especially considering all these bugs I'm encountering are None type errors.

The smartslice result object gets initialized with the status as I see here, so unless `.getResult()` has some weird behavior I'm not aware of I think this is solid.
https://github.com/tetonsim/pywim/blob/be6eddab85d08c414f1f544b0584a87fcdc9bc80/pywim/smartslice/result.py#L39-L45

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -208,16 +208,22 @@ def _saveState(self, writing_workspace=False):
         self._storage.setEntryToStore(plugin_id=self.metadata.id, key="status", data=status.value)
 
         # Need to do some checks to see if we've stored the results for the active job
-        if cloudJob and cloudJob.getResult():
-            self._storage.setEntryToStore(plugin_id=self.metadata.id, key="results", data=cloudJob.getResult().to_dict())
+        if cloudJob and not cloudJob.getResult().status is pywim.smartslice.result.ResultStatus.no_solution_found:
```
dougkenik, line 211: Is it possible for `cloudJob.getResult()` to be `None` here?

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -208,16 +208,22 @@ def _saveState(self, writing_workspace=False):
...
+            self._storage.setEntryToStore(
+                plugin_id=self.metadata.id,
+                key="results",
+                data=cloudJob.getResult().to_dict()
+            )
             self._storage.setEntryToStore(
```
dougkenik, line 217: I'm a little fuzzy here.... If this is a validation, and not an optimization, what is it storing for `self.proxy.resultsTable.getSelectedResultId()` ? There is technically nothing selected, because there is only 1 result

dougkenik (changes requested):

SmartSlicePlugin/SmartSliceExtension.py
AijunHall: Not sure 100%, but since that was a previous check, I will include it back in.

SmartSlicePlugin/SmartSliceExtension.py
AijunHall: This is previously existing code I just shifted so I can't be much of an authority here.

I'm doing some testing, so far and it seems to be storing 0. It does this for validations AND optimizations actually.
`self.proxy.resultsTable.getSelectedResultId():  0`

Is this deprecated?

dougkenik: No - select a different row in the optimization table and save that. The value should change to that row

AijunHall: Heres the reference in the repo I found if this helps either of us:

https://github.com/tetonsim/is-cura-ui/blob/2b2b71cc8dcb3a42fd7527358459acc0b9eac8e1/SmartSlicePlugin/SmartSliceCloudProxy.py#L789-L791

dougkenik (approved):
Looks good. We will need to test it thoroughly

bradymadams (approved):

bradymadams:
Will merge pending approval from @jake-kirkland-teton

jake-kirkland-teton (approved):

PR #383, opened 2021-01-27
R20.2 bug 374

### LINKED ISSUE #374: Smartslice stage error overlapping
See GIF. When a user has the error message box up, and changes a setting to an invalid one, the error message box won't extend in realtime. It only corrects once the user 'refreshes' the error message box by hovering over it again.

This can be reproduced a number of different ways - wherever a setting can be changed in one click with the error message box open at the same time basically.

[Smartslice error overlapping](url)

dougkenik:
Fixes #374


jake-kirkland-teton (changes requested):
There is still one situation that causes this bug. If you change a print setting while the validate button is "sick" and you make it acceptable, the validate button becomes active but the resolved window remains.
Bug374

dougkenik:
Good to go now

AijunHall (approved):
Been trying to reproduce the error overlapping for a while now, and haven't been able to do it. Looks like you've got it fixed 100% on the user end.

SmartSlicePlugin/stage/ui/SmartSliceMain.qml
```diff
@@ -849,11 +849,26 @@ Item {
...
 
+                    Connections {
+                        target: smartSliceMain.proxy
+                        onSmartSliceErrorsChanged: {
+                            smartSliceErrors.forceLayout();
+                            smartSliceWarningContents.forceLayout();
+
+                            if (errorRepeater.model.length == 0 && smartSliceWarningPopup.opened) {
+                                smartSliceWarningPopup.close();
+                                return
```
bradymadams, line 867: This return can be removed, but probably not worth the effort unless you need to make some other changes.

bradymadams (approved):

jake-kirkland-teton (approved):

PR #386, opened 2021-01-27
R20.2 account link

dougkenik:
- Added an "account" URL to plugin.json
- Added an "Account" option in the Smart Slice extension
- Changed the forgot password link to us the new account URL

I'd be interested to hear opinions on how this was architected. I had to get the plugin metadata from plugin.json over to the SmartSliceProxy in order to use it in the QML code, and I did so by just sending the metadata in the constructor for the proxy.


bradymadams (approved):

colman-hartinger (approved):

jake-kirkland-teton (approved):
I think this is actually a decent change, I think it will be very helpful in the future if we need the metadata for another reason, it will be accessible from `proxy` which is a good place to have it.

PR #387, opened 2021-01-29
R20.2 load direction incorrect

### LINKED ISSUE #367: Load direction is incorrect after opening saved model
The load direction is incorrect when opening a saved model. The steps to reproduce are shown in the attached workflow video but are briefly summarized here:

1. Load the attached STL file.
2. Rotate the part on the build plate as shown in the video. This step is critical. I am not able to reproduce this if I do not rotate the part.
3. Setup the loading as shown in the video. Note that the load direction is normal to the surface.
4. Run a validation. This step is not critical. The bug occurs if you save before running validate.
5. Save the project.
6. Open the project and view the loads. The load direction will now be tangent to the surface instead of normal. This is the bug.

Based on the above, it appears this issue is somehow related to the rotation of the part.

[load_direction_bug.zip](https://github.com/tetonsim/is-cura-ui/files/5706026/load_direction_bug.zip)

jake-kirkland-teton:
This was painful to figure out, but I think it is solved in the way of load arrows. Please test and let me know if you find a configuration that still causes this bug.

Closes #367


bradymadams (commented):
This looks good to me. I just have some thoughts about minor changes to the function interfaces.

1. In `createSteps` could we rename `writing_workspace` to something more indicative of how it changes the logic, like `transform_bcs`?
2. Can we remove the `writing_workspace` argument from the `pywimBoundaryCondition` functions and just have `createSteps` control the rotation of the bcs by modifying the `mesh_rotation` argument appropriately? (Identity matrix if `transform_bcs` is false).

Something like that just seems a little cleaner for future devs to read and understand how the functions work.

dougkenik (changes requested):
Ran the attached job, saved it, and then tried to open it. Cura crashes (rename to .3mf)

[UMS5_pedal_fixture-v2_z_build.zip](https://github.com/tetonsim/is-cura-ui/files/5894399/UMS5_pedal_fixture-v2_z_build.zip)

jake-kirkland-teton:
@bradymadams So I can absolutely rename that variable, that is a good call.

On the second point, the problem is, we do not want to adjust the bcs by `cura_to_print` either, and if `mesh_rotation` is passed in as identity, that works for the rotation, but the bcs will still be transformed by `cura_to_print`. I could rewrite the check so see if `mesh_rotation` is identity, and then prevent that transformation from there, but that might also cause unwanted issues if the rotation matrix is in fact identity and we DO want to use `cura_to_print`. I like the idea, I just don't think it will work correctly, unless I am over thinking it?

jake-kirkland-teton:
That is fixed now @dougkenik, I was accessing the object incorrectly after I shuffled the types.

dougkenik:
Ok - I would mark this as something that needs a unit test then :). That should have been caught easily by one of those

dougkenik (approved):
Much better

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -262,17 +262,34 @@ def setMeshDataFromPywimTriangles(
...
+        force_vec = [load_vec.x, load_vec.y, load_vec.z]
+
+        # To prevent incorrect load arrow restoring from file, we have to deal with the Coordinate System a little differently.
+        # If we are NOT saving the project to a file, we want to rotate the loads so they can be calculated by the back end
+        #   when they are sent in.
+        # If we ARE saving the project, we do not want to rotate the loads, since the faces are not rotated to match on the mesh.
+        #
+        # When we do rotate the mesh, we first rotate it according to the rotation of the part on the printer stage,
+        #   then we do the second rotation to change it to the coordinate system for the back end (using the cura_to_print matrix)
+        if not transform_bcs:
```
bradymadams, line 278: Shouldn't this be `if transform_bcs`?

jake-kirkland-teton: Actually it should not. We only want to transform the BC's if we are submitting a job to the back end, not when we are saving. So when this is `True` we are saving the project, and we *do not* want to transform the BC's. Which means we should probably rename that again. I wasn't thinking clearly enough about it.

colman-hartinger (approved):
code looks good and it doesn't seem like the load will switch now

SmartSlicePlugin/stage/SmartSliceScene.py
jake-kirkland-teton: Would `do_not_transform_bcs` work?

bradymadams: It would but a double negative here seems unnecessary and makes the code harder to read. Just use `if transform_bcs` and when `createSteps` is called: `createSteps(transform_bcs=not writing_workspace)`.

jake-kirkland-teton: Ahhh yes. That is a good way to do it. I'll change that real fast.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -188,7 +188,7 @@ def disableTools(self):
 class AnchorFace(HighlightFace):
     color = Color(1., 0.4, 0.4, 1.)
 
-    def pywimBoundaryCondition(self, step: pywim.chop.model.Step, mesh_rotation: Matrix):
+    def pywimBoundaryCondition(self, step: pywim.chop.model.Step, mesh_rotation: Matrix, transform_bcs=False):
```
bradymadams, line 191: This was originally defaulted to `writing_workspace=False` which meant that any calls that didn't set this argument would transform the BC. So should this be changed to `transform_bcs=True` now that the `if` is swapped?

jake-kirkland-teton: Honestly I was defaulting it to the more likely case, since saving happens more rarely than validating/checking the BC's, and leaving it `False` now will default to executing less code. Realistically, it is only in place as a backup, and the one for `Anchors` is just being extra safe since it is never actually used. 

That all being said, I think defaulting to `True` is probably still better just in case something goes awry and it isn't sent down the pipe properly. I will change it real quick. Good catch.

bradymadams: Well, in a case like this you also need to consider existing code. I'm not sure if all the calls to `pywimBoundaryCondition` got updated in this PR, but if they didn't then you have to consider what those untouched calls are expecting. It seems to me they'd be expecting the BC to transform since that was the original behavior. So I think defaulting it to True is the right approach.

bradymadams (approved):

PR #390, opened 2021-02-02
Fix for the displacement rotation issue

### LINKED ISSUE #388: Displacement Mesh CSYS Bug
Similarly to Issue #367, we are seeing the displacement mesh displacing to the incorrect axis if the part is rotated in certain ways.

jake-kirkland-teton:
Correctly transformed and rotated the displacements for display. Added supporting comments.
Huge thanks to @jlselden for all the help getting this one sorted out.

Closes #388


jake-kirkland-teton:
@jlselden Hey Jeff, I looked into this one, and we are actually getting a [0, 0, 0] displacement from the back end for the points in this mesh that aren't moving. Any ideas?

jake-kirkland-teton:
In the following screenshot it is:
[ID] [Unmodified displacement] [CSYS swapped displacement] [Unrotated displacement]

image

jlselden:
I'll take a look.

jlselden:
@jake-kirkland-teton @dougkenik This is happening when the displacements get mapped to the STL mesh on the backend. If a vertex in the STL mesh tries to reference a nonexistent voxel, it defaults to a displacement value of 0 and spits out a message. That's what is happening in this case, although I still need to look into why the reference voxel doesn't exist. If I can't figure it out fairly quickly, I'll pull Faisal in since this part was written by him.

jlselden:
In addition to incorrectly fixed points, it also looks like the part is flexing the wrong direction in y. It should be flexing slightly towards the front of the plate. Might need to recheck the negative sign on the y.

jlselden:
Just submitted a PR to add some neighboring voxels that were missing from the search list. This issue would have shown up the more we tested, so hopefully, we've nipped it in the bud.

bradymadams:
Just deployed a new build to the test server containing Jeff's fix.

jake-kirkland-teton:
Looks like that fixed the problem @jlselden. I suppose we will find out through the team testing if it reoccurs.

bradymadams (approved):

dougkenik (approved):

Issue #377, opened 2021-01-20
Analysis failed with custom print settings


dougkenik:
Running a job in Cura that has some customized settings. One (or more) of the customized settings is causing the analysis to fail when it's submitted. We should walk through each custom setting to see what is causing the issue.

![image.png](https://images.zenhubusercontent.com/5ef3733ee554441fdbf0d022/e3af3c26-e803-4ff3-a9ee-14160c79fb04)

AijunHall:
Its the **Support** and **Build Plate Adhesion** Extruder settings. This specific analysis failure error comes up when these settings are not set to Extruder 1.

![image](https://user-images.githubusercontent.com/54338983/106328868-614a9a80-6235-11eb-8ab8-12e4a7b6fe0a.png)

![image](https://user-images.githubusercontent.com/54338983/106329397-314fc700-6236-11eb-9776-e2e93493382d.png)

PR #392, opened 2021-02-02
fix for bug 377

dougkenik:
Fix for #377 . We now track all of the support extruders and the adhesion extruders.


bradymadams (approved):

colman-hartinger (approved):

jake-kirkland-teton (approved):

bradymadams:
@dougkenik what does this change mean for the status of #375?

dougkenik:
When I open the file, Validate goes sick and asks to change extruders. However, if you turn off "Enable Support" now, it will still check the support extruders. We will need to create another ticket to not check the support Extruder Keys if support is disabled. 

The job errors out gracefully now due to the voxel sizes

PR #394, opened 2021-02-03
Added checks for disabled Support and Adhesion

### LINKED ISSUE #393: Don't check support extruders if support is not enabled
We should not be checking the support extruders if support is disabled. This was introduced in #399

jake-kirkland-teton:
If support extruders or adhesion are disabled, we will no longer be checking them when properties change.

I believe this is fixed as requested. If I missed something please let me know, but we are now using similar checks to Cura in the `ExtruderManager.py`

Closes #393


SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -164,8 +165,8 @@ def buildSceneNode(self, node):
         scene_node = SmartSliceProperty.SceneNode(node, node.getName())
         self._properties.append(scene_node)
         Logger.log("d", "Tracking properties for {}".format(node.getName()))
-        stack = node.callDecoration('getStack')
-        stack.propertyChanged.connect(self._onSceneNodePropertyChanged)
+        self._stack = node.callDecoration('getStack')
```
dougkenik, line 168: I don't think this is what you want.... you are storing stack at the SmartSlicePropertyHandler level for a given node. What happens. If a user selects a modifier mesh and changes something?

SmartSlicePlugin/SmartSlicePropertyHandler.py
```diff
@@ -305,9 +306,22 @@ def _onSceneNodeChanged(self, node=None):
...
-        if key not in SmartSliceProperty.ExtruderProperty.NAMES + SmartSliceProperty.ExtruderProperty.EXTRUDER_KEYS:
+        if key not in (
+            SmartSliceProperty.ExtruderProperty.NAMES +
+            SmartSliceProperty.ExtruderProperty.EXTRUDER_KEYS +
+            SmartSliceProperty.ExtruderProperty.ADHESION_EXTRUDER_KEYS +
+            SmartSliceProperty.ExtruderProperty.SUPPORT_EXTRUDER_KEYS
+        ):
             return
 
+        if key in (SmartSliceProperty.ExtruderProperty.SUPPORT_EXTRUDER_KEYS):
```
dougkenik, line 317: I think this is the wrong place to put these checks

dougkenik (changes requested):
I don't think the checks you put in are at the right level. I would suggest modifying the class `SceneNodeExtruder`. It will be much cleaner because you can check the stack of the node in question instead of assuming the stack associated with any given node.

As it is now, you would change `self._stack` in `SmartSlicePropertyHandler` any time you created a scene node (normal mesh, modifier mesh, etc.).

jake-kirkland-teton:
Got that changed.

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -202,14 +208,24 @@ def __init__(self, node=None):
...
         if self._node:
+            stack = self._node.callDecoration("getStack")
             active_extruder = getNodeActiveExtruder(self._node)
 
             active_extruder_index = int(active_extruder.getMetaDataEntry("position"))
 
             specific_indices = {}
-            for key in ExtruderProperty.EXTRUDER_KEYS:
+
+            key_list = ExtruderProperty.EXTRUDER_KEYS
```
dougkenik, line 221: Can we create a private function to this class for acquiring 'stack' generating 'key_list'? It seems to be used in multiple places

dougkenik (changes requested):
One more request. I'll test functionality after this.

SmartSlicePlugin/SmartSliceProperty.py
jake-kirkland-teton: Very fair point. Got that added.

dougkenik (changes requested):
The code looks better, but it doesn't seem to be working?

image

jake-kirkland-teton:
Alright, so lets take a step back. What exactly is supposed to be fixed in this bug fix? What is the exact issue? How do I replicate the bug? 

As of right now, I am preventing the extruders/adhesion from getting checked by our system if they are disabled. If you had them enabled, or if you enabled them and disabled them, we were getting a crash. I was under the assumption that that bug was the one we wanted to fix, based on the description of the bug.

jake-kirkland-teton:
Ah, I think I see what you mean. I'll look into it. We are attempting to prevent extruder 2 on the support extruder as well correct?

dougkenik:
> Alright, so lets take a step back. What exactly is supposed to be fixed in this bug fix? What is the exact issue? How do I replicate the bug?
> 
> As of right now, I am preventing the extruders/adhesion from getting checked by our system if they are disabled. If you had them enabled, or if you enabled them and disabled them, we were getting a crash. I was under the assumption that that bug was the one we wanted to fix, based on the description of the bug.

I am uncertain about what your statement is concerning. Here is what was happening:

I had put in a fix to check the support extruder number and the adhesion extruder number and add them to the error list if they were not Extruder 1 (like we do with infill extruder, etc). After the changes were merged, I found an issue that we were still checking the support and adhesion extruder number if support and adhesion were **not** turned on. This is not a good user experience, because if they are not turned on, we should not be checking them. 

When I'm testing your PR, I can turn on support, and then change the extruder number to 2. I'm happy to walk you through this in a call

dougkenik:
> Ah, I think I see what you mean. I'll look into it. We are attempting to prevent extruder 2 on the support extruder as well correct?

If support is enabled, we should be issuing an error when extruder 2 is selected for any support extruder. Same for the adhesion extruder

jake-kirkland-teton:
Alright @dougkenik :

I got the check fixed for extruder 2. I didn't realize that the validate "sickness" was using the `EXTRUDER_KEYS` list. The only problem with that, is we have to rebuild the list of extruder keys there too, unless you know of a way I can get a `SceneNodeExtruder` object relating to the node being checked that I can pass in?

Secondly, changing the support/adhesion enabled status does not cause a signal to hit `checkJob` to recheck if the validate button should enable/disable. What signal do you think would be the best to hook it up to? I can hook it to a change in extruder settings, but it seems like it will hit `checkJob` a lot if I do.s

dougkenik:
Well... SmartSlicePropertyHandler has a list of SmartSliceProperty.SceneNode, which has, in turn, SceneNodeExtruder and a SceneNode

If you have access to SmartSlicePropertyHandler  in that function, you could create a routine to get a SceneNodeExtruder by comparing the node, then just use the function you have already created to generate the list of keys.

Dunno if that is worth it or not. Think about it and let me know

jake-kirkland-teton:
I was wondering if that was how those were accessed. Hmm. I can absolutely do it that way, though at this point I think the 4 lines I wrote that are a slight duplication should be a better solution overall. Since the `SceneNode` objects are appended to the `_properties` list, we would have to use a function to go through all of the `_properties` to find the SceneNodes, then iterate through the SceneNodes to find the one in question, then again iterate through the extruder properties for that SceneNode to get the one we want. This will cut a lot of iterating out of the process. I'm open to discussion.

How about activating `checkJob` when the support/adhesion settings are changed? Any thoughts on that issue?

dougkenik:
Wouldn't `checkJob` be called when they are changed the way it is, but only if support and/or adhesion are active?

jake-kirkland-teton:
Actually I think it just occurred to me how to fix it easily. But yes and no. If support/adhesion IS enabled, changing some of the settings will absolutely cause `checkJob` to occur, but the actual changing of support enabled and adhesion enabled will not. Disabling adhesion enabled will, enabling it will not. Neither will on support enabled.

jake-kirkland-teton:
Hmmm. OK, not as easy as I had hoped. I forgot we are pulling in all of the values from these keys as `int` and `adhesion_type` is a `string`/`NoneType`

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -113,7 +114,12 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_extrude
             job.chop.meshes.add(mesh)
 
             # Check the active extruder
-            any_individual_extruder = all(map(lambda k : (int(active_extruder.getProperty(k, "value")) <= 0), ExtruderProperty.EXTRUDER_KEYS))
+            extruder_keys = ExtruderProperty.EXTRUDER_KEYS
+            if node.callDecoration("getStack").getProperty("support_enable", "value"):
+                extruder_keys = extruder_keys + ExtruderProperty.SUPPORT_EXTRUDER_KEYS
+            if CuraApplication.getInstance().getGlobalContainerStack().getProperty("adhesion_type", "value") != "none":
+                extruder_keys = extruder_keys + ["adhesion_extruder_nr"]
```
dougkenik, line 121: I would put the adhesion extruder in the ExtruderProperty class just so we can keep code consolidated

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -75,7 +75,9 @@ class GlobalProperty(ContainerProperty):
         "quality",
         "magic_spiralize",
         "wireframe_enabled",
-        "adaptive_layer_height_enabled"
+        "adaptive_layer_height_enabled",
+        "adhesion_type",                      # Adhesion type toggle
+        "adhesion_extruder_nr"                # Build plate adhesion extruder
```
dougkenik, line 80: See previous comment about putting the adhesion extruder in a different property so we can keep track of extruders, and properties

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -106,12 +108,18 @@ class ExtruderProperty(ContainerProperty):
...
         "support_infill_extruder_nr",       # Support infill extruder
         "support_extruder_nr_layer_0",      # Layer 0 extruder
         "support_bottom_extruder_nr",       # Support bottom extruder
-        "support_roof_extruder_nr",         # Support roof extruder
-        "adhesion_extruder_nr"              # Build plate adhesion extruder
+        "support_roof_extruder_nr"          # Support roof extruder
+    ]
+
+    SUPPORT_ENABLED_KEYS = [
+        "support_enable"                   # Support enabled toggle
```
dougkenik, line 122: If you are going to have this, you should use it in SmartSliceJobHandler, and you should also have the adhesion property separated in a different key

dougkenik (changes requested):

SmartSlicePlugin/SmartSliceJobHandler.py
jake-kirkland-teton: So the problem with putting the adhesion extruder in the extruder property class, is that it is a global property, not an extrusion property. So when it is in extruder property, it does not actually track it, because it doesn't come from the ExtruderStack. Which is why I placed them both in the global property class to be tracked. And since they are now in global properties, it would take a substantial rewrite of how those properties work to make them more similar to the extruder properties if we want to split them off into their own list. Additionally, we don't want adhesion_type to go along with the properties a lot of the time, since we only want to track it to know when it changed, we don't actually care what the value is, and it doesn't help us in the check to enable/disable the validate button. If we wanted to include that in the list as well, we would have to rewrite our checks to not type cast things into int, we would have to split that one property off in the checks for extruder 1, etc.

SmartSlicePlugin/SmartSliceProperty.py
jake-kirkland-teton: The reason this one is split off, is because we do not want it in the SmartSliceJobHandler check, because when support is enabled, but everything else is extruder 1 and acceptable, the way the check is written is making sure nothing is greater than one. When you include `support_enable` and you have support enabled, but it is extruder one, you throw a false error since the check finds a 1 rather than all 0s and tells you your extruder needs changed.

SmartSlicePlugin/SmartSliceJobHandler.py
dougkenik: Ok. That comment lost me. Can we put the variable in the SmartSliceProperty.GlobalProperty class, as a different key?

jake-kirkland-teton: Yeah. I think I can do that without major changes. The key will just be listed in both `NAMES` and the new key list. That ok?

jake-kirkland-teton:
OK. I should have this fixed @dougkenik, let me know what you think. 

@bradymadams On lines 119 and 121 in `SmartSliceJobHandler` the lines are getting really long, but I am not exactly sure how to split them into multiple lines. How would you recommend that be done?

bradymadams:
> @bradymadams On lines 119 and 121 in `SmartSliceJobHandler` the lines are getting really long, but I am not exactly sure how to split them into multiple lines. How would you recommend that be done?

I would pull the string of method calls out into a variable. This will shorten the line and also improve efficiency. The way it is - those call strings will get executed every time the lambda is called as `any` iterates.

There's a lot going on that part of the code - I'd also add some vertical whitespace to improve readability.

dougkenik (approved):
Functionality looks great now. I like the code cleanup too

colman-hartinger (approved):
code looks good and the bug seems to be fixed as well

PR #424, opened 2021-02-09
Expand check for first time use (404)

### LINKED ISSUE #404: All BC faces highlighted on project open + erroneous highlighted face
Steps to repro:
1. [Open this 3MF in Cura](https://github.com/tetonsim/is-cura-ui/files/5945922/bracket-highlighted-faces.3mf.zip)
2. Open Smart Slice

Notice that all of the BC faces are highlighted, which isn't expected. This isn't too bad as the highlighting still behaves well as you use the tool. But also notice the erroneously highlighted bottom face. I'm not sure if these two issues are related, but I'll group them together in this one issue until we find out more.

bracket-highlighted-faces

bradymadams:
Fixes #404. Turns out this was an artifact of logging in to Cura from source. The installed Cura picked up the token correctly, but the username wasn't in the preferences. So I simply expanded the check to also check if the user is logged in. This will correctly catch non first time users that have multiple Cura installs and users that have an expired token.


dougkenik (approved):
Great catch

jake-kirkland-teton (approved):

PR #427, opened 2021-02-10
Fix for modifier mesh restoring.

### LINKED ISSUE #418: Modifier Mesh Is Moved Below the Stage On Cancel Changes
Steps to reproduce:

1) Load an optimized project that has a modifier mesh.
2) On the prepare stage, delete the modifier mesh.
3) Cancel your changes.

[https://app.zenhub.com/files/209108148/3cff0763-5d60-44c7-8b00-6bf504940733/download](https://app.zenhub.com/files/209108148/3cff0763-5d60-44c7-8b00-6bf504940733/download)

jake-kirkland-teton:
We were restoring the modifier meshes as a child of the build plate rather than the printable model, which was causing translation issues due to different origin points for the coordinates. I have modified that so now the modifier mesh is reattached correctly to the printable node.

Closes #418


SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -415,13 +415,19 @@ def restore(self):
             if n in nodes:
                 continue
 
-            self._root.addChild(n)
+            if isinstance(n, SmartSliceMeshNode):
+                self._nodes[0].addChild(n)
```
dougkenik, line 419: I don't think this is safe, or extendable. We should be calling node.getParent() instead of using self._nodes[0]. However, that will cause a problem (see me general comment as to why)

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -415,13 +415,19 @@ def restore(self):
...
+            else:
+                self._root.addChild(n)
 
         for n in nodes:
             if n in self._nodes:
                 continue
 
-            self._root.removeChild(n)
+            if isinstance(n, SmartSliceMeshNode):
```
dougkenik, line 427: See above comment on this

dougkenik (changes requested):
I think this is a little "hacky." When I was looking into this, the parent of the node is getting set to None somewhere when it is deleted (when it should be the printable node). We should be trying to figure out what is causing the parent to become None for that node, and then tap into that signal.

PR #436, opened 2021-02-11
Fix duplicate mod meshes on project open

### LINKED ISSUE #431: Mod meshes are duplicated and translated when opening .3mf from SS version 20.1
Steps to reproduce:

1. Install SS 20.1.
2. Run an optimization. Save the project.
3. Install 20.2 release candidate 1 or 2.
4. Open the .3mf from step 2.
5. The modifier meshes from the optimization will be (a) duplicated and (b) the duplicate set will be translated.

Expected: the mod meshes are in the same position as they were in 20.1.

The following screenshots show the optimized part from step 2 (SmartSlicePlugin-r20.1.103) and the optimized part from step 4 (SmartSlicePlugin-r20.2.44). The 2nd image clearly shows that the mod meshes were copied and translated.

Optimized .3mf from 20.1:
[optimized_side_orientation.zip](https://github.com/tetonsim/is-cura-ui/files/5962367/optimized_side_orientation.zip)

image

image

bradymadams:
Fixes #431 

The duplication was removed. The translation was due to the setting that Cura uses to move meshes to the plate automatically. Fortunately, that translation was performed on the meshes that Cura was adding. Those are now removed, so the single mod mesh left is in the correct location.

[Build](https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/SmartSlicePlugin-r20.2-duplicate-mod-meshes.45.curapackage)


dougkenik (approved):
Very Nice

jake-kirkland-teton (approved):

rickdalgarno (approved):
Tested the workflow in #431 and the issue is fixed.

PR #437, opened 2021-02-11
R20.2 429

### LINKED ISSUE #429: Failure region popup text should not change when the user changes the FoS.
Currently, a user can plot the failure regions. The regions that are plotted depend on the FoS value entered by the user when the job was submitted. If the user plots the failure regions, then changes the FoS value and replots the failure regions, the failure regions do not change. However, the popup text does change and it implies the region now corresponds to the updated FoS.

Workflow:

1. Define a FoS of, for example, 1.5.
2. Run a validation.
2. Plot the failure regions. Note that the popup says "The areas shaded in red have... value of 1.5".
3. Change the FoS to, for example, 1.3.
4. Plot the failure regions again. Note that the failure regions are the same as from step 2 but the popup says "The areas shaded in red have... value of 1.3.

Expected behavior: When the user changes the FoS, the popup should continue to display "The areas shaded in red have... value of _original FoS_".

Important: We don't want to reset the validation results when a user changes the FoS. It is common for a user to run a validation, THEN define the FoS for the optimization.

Config: Cura 4.7 and release candidate 2.

dougkenik:
Now require a revalidation if the requirements are changed after validation or during a validation. This is because the problem meshes change depending on the factor of safety requirement. Fixes #429


rickdalgarno:
@dougkenik build?

rickdalgarno:
Testing this reveals a new issue. If the reported FoS is less than 1.0, a user may want to set the target FoS to a value < 1.0. Currently, we don't allow users to enter FoS < 1.0.

The use case here is a user will want to plot regions where the FoS is < 1.0 so they they can pin-point the trouble areas.

dougkenik:
@rickdalgarno - can you create an enhancement?

rickdalgarno:
> @rickdalgarno - can you create an enhancement?

Created in #439

rickdalgarno (approved):
Tested using SmartSlicePlugin-r20.2-429.46 and confirmed that a revalidation is triggered when (a) the requirements are changing during a validation and (b) after a validation.

jake-kirkland-teton (approved):

bradymadams (approved):

PR #440, opened 2021-02-12
Added stack for controlling property change lock

### LINKED ISSUE #434: Opening the 3MF with optimized results causes the "cancel or continue" dialog to show
When I open this 3MF, the popup displays about losing results. That should not happen when opening an optimized 3MF[https://app.zenhub.com/files/209108148/895aef08-41e6-4224-a30a-d9d299e2348c/download](https://app.zenhub.com/files/209108148/895aef08-41e6-4224-a30a-d9d299e2348c/download)

jake-kirkland-teton:
Closes #434


bradymadams (changes requested):
1. The `_cancelChanges` attribute is still present in the PropertyHandler class, and there is remaining logic that is trying to use it (see function `cancelChanges`). This probably needs to be changed to use the new functionality.
2. When we talked about this on Slack I pointed out that `_addProperties` looked like it was basically another way of using `_cancelChanges` and that we should fold this into the new functionality. Did you run into a problem with trying to replace this attribute with the pause/resume functionality?
3. I'm worried about the usage of pause in `SmartSliceCloudConnector.submitSmartSliceJob`. How positive are you that resume will be called somewhere else if pause is called in this function? The way I pictured this functionality being used was pause and resume would essentially be called right next to each other so that it wasn't confusing to understand or maintain how the property tracking was being controlled. I didn't spend a lot of time following the code paths, but it definitely was not obvious to me where the tracking gets resumed if it is paused in `submitSmartSliceJob`. I'd even think about it like using `with` statement (in fact, implementing this so a `with` is used would be the safest approach because it would protect against programmer errors and caught exceptions. I didn't mention it earlier because it would be a little extra work, but it might help to at least think about it in those terms).

```python
with propertyHandler.pauseTracking():
  # do stuff
# when the with exits our pause request automatically gets popped from the stack
```

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -230,6 +230,8 @@ def _saveState(self, writing_workspace=False):
                 if writing_workspace:
                     cloudJob.saved = True
 
+            self.cloud.propertyHandler.resumeTracking()
```
bradymadams, line 233: Looks like this might be the resume for the other pause in `submitSmartSliceJob`, but it's hard to follow. Which means this is going to be prone to programmer error in the future.

bradymadams (changes requested):
On a closer look I see that in `SmartSliceCloudJob.run` you check for the task being None and if so resume tracking which should revert the pause request done in submitSmartSliceJob. However, there are two pauses in submitSmartSliceJob.

I understand you are basically trying to replace the places in the code that use to set `_cancelChanges`, but I think it might be more involved then that. I was hoping we could clean up all the old code and replace it with very obvious pause/resume pairs. If that doesn't seem possible let's take a step back.

jake-kirkland-teton:
1) You are absolutely correct.That's my fault. I should have cleaned that up better.
2) Apologies. I think I was in the "quick fix" mindset and I misinterpreted what you said. If we decide to continue with this method, that should be an easy addition.
3) The clearer pause and resume certainly makes sense to me. I can try to clear it up some, we may be able to consolidate all of the pauses and just set them to one single pause right before the properties are changed, rather than the way it is now. I am not positive, I feel like the way it was setup currently was to prevent tracking during job submission.
On that point, the two calls to pause do in fact both get resumed, as far as I can tell, always. You are correct in that the result calls either get resumed after a None is returned, or after the results are set from the backend results.

So the question, is: Is it worth it at this point to keep working on this to make it cleaner and more understandable, or do we want to change direction?

bradymadams:
> So the question, is: Is it worth it at this point to keep working on this to make it cleaner and more understandable, or do we want to change direction?

I'm not sure if I feel ready to make a call on that without taking a deeper look. Let me know if you have an opinion on possible approaches.

jake-kirkland-teton:
Well, the way I see things, is we need this fixed for the release right now. So with that in mind, I have come up with a few options:

1) I can revert the job tracking stuff, but leave the pause/resume and stack for just the fix on this bug for the time being, with a "TODO" left to reconvert and test more clearly the existing system switch.

2) I can test out a theory I have that we may not need to pause the tracking in those locations before the job results are returned, and we really only need to pause the tracking on a successful job when the results will be pulled out of it. Though there may be some hidden pitfalls I am not currently aware of/seeing that led to us pausing the tracking at those times in the first place.

3) We can come up with a totally new way to solve this bug and revisit the entire system later.

I am happy to attempt any of these fixes, but I really do like the concept of the stack system. I think that was a smart idea and with correct implementation it will be an elegant system in place to help clarify and fix this issue, and be modifiable in the future. I also agree, that this is not exactly a clear system, and implementing this the complete correct way may require more time than we have at this point.

bradymadams (approved):
Tested opening an optimized 3MF and the change dialog no longer opens. I tested various workflows with the optimized 3MF and everything behaved as expected.

dougkenik (approved):

Issue #405, opened 2021-02-08
Deformed shaped and Failure locations pop-up boxes text formatting


rickdalgarno:
The pop-ups for the deformed shape and failure locations tools should be cleaned up in my opinion. 2 issues to address are provided in the annotated images below.

![pop-up-suggestions](https://user-images.githubusercontent.com/34459028/107266918-6c56b500-6a03-11eb-9a96-95783d40459e.png)

faisal-bhuiyan:
Same observation for the following. We need to clean up the formatting of the text in the displayed message.

![Untitled](https://user-images.githubusercontent.com/59463700/107271373-5cda6a80-6a09-11eb-8415-eae72556b0bf.png)

dougkenik:
I think this is machine and/or display dependent. It does not display this way display. We need to fix it

PR #442, opened 2021-02-12
R20.2 405

dougkenik:
Brady figured out how to fix the formatting, so I implemented it.

I also fixed the material warning popup. It was displaying at odd times, so I changed the behavior. It will only show the first time a user selects a material which has not been tested, but uses a generic equivalent


bradymadams:
[Build](https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/SmartSlicePlugin-r20.2-405.48.curapackage)

bradymadams (approved):

rickdalgarno (changes requested):
Tested and confirmed fix for #405.

However, there is no material popup when I select an untested material. I cycled through several untested materials, and never saw a "this material has not been tested.. generic equivalent" popup.

rickdalgarno (approved):
Choosing a variety of tested, supported and not tested, and unsupported materials results in expected behavior.

jake-kirkland-teton (approved):
I'm both glad and pissed at how easy this change was. Nice job @bradymadams

PR #463, opened 2021-02-22
unit test fix

colman-hartinger:
Still testing to make sure that action will fail when unit test fails


build_unit_tests.sh
```diff
@@ -18,4 +18,7 @@ mv ${CURA_APP_DIR}/materials ${CURA_APP_DIR}/Cura/resources/materials
 mv SmartSlicePlugin ${CURA_APP_DIR}/Cura/plugins
 
 apt-get install -y mesa-utils xvfb
+git clone https://github.com/tetonsim/pywim.git
+mv pywim/pywim /srv/cura/Cura/plugins/SmartSlicePlugin/3rd-party/cpython-common
```
bradymadams, line 22: Why is this needed when it wasn't before?

bradymadams (changes requested):
In build_unit_tests.sh there is:

`git clone -b ${1} --depth 1 https://github.com/Ultimaker/dm_materials materials`

That repo doesn't exist. Should it be fixed? Or is it even necessary since the tests seem to pass without the intended fdm_materials repo?

build_unit_tests.sh
colman-hartinger: we were just using a pip install for pywim before but that seemed to be an old version, If you loo'k at [this](https://github.com/tetonsim/is-cura-ui/runs/1956109651?check_suite_focus=true) action that failed it was because pywim didn't have the 'ModelRegion' attribute and this fixed that

bradymadams: I actually just published the new pywim about an hour ago, so `pip` should work now. I'm fine with it cloning from github as well, but it should probably checkout a specific branch. `master` will usually lag.

colman-hartinger: I think it's smoother just to do a pip install so I removed those lines. Should I note in the README that if tests are failing to change to the latest branch of pywim?

bradymadams: Let's leave it out for now. I'd like to think about that and how we might be able to more reliably target the right branches of the external dependencies.

bradymadams:
@colman-hartinger did you see my other question about the fdm_materials repo?

colman-hartinger:
> @colman-hartinger did you see my other question about the fdm_materials repo?

Totally missed that, just fixed the typo. Although not necessary right now It still seems like a good idea to have that in case we want to test material changes

bradymadams (approved):

Issue #465, opened 2021-02-23
Retrieving models for Unit Tests


bradymadams:
Something we'll need to consider here is whether or not we want the objects in S3 to be publicly available. If they're not we'll need to make sure we don't publish any access keys in our github action.

PR #473, opened 2021-02-26
R21.0 unit test state management

colman-hartinger:
Working state management system for SmartSlice unit testing

I wasn't able to bypass the dialogs that popup when opening a 3mf so instead I implemented a system where an stl is loaded in and then the corresponding smart slice data is loaded in afterwards. Any changes outside the test folder facilitate that system.

Also I was getting blocked trying to implement a changePrinter function. Looking through relevant code in Cura and UM it seemed that any class referencing a printer was looking for a printer on the network.  @bradymadams If you could point me in the direction of where you think the actions are to add a non-networked printer and then change the printer, I should get that done in no time.


SmartSlicePlugin/select_tool/LoadToolHandle.py
```diff
@@ -50,19 +50,20 @@ def buildMesh(self):
...
-        self._all_axis_color = Color(*theme.getColor("all_axis").getRgb())
-
-        self._axis_color_map = {
-            self.NoAxis: self._disabled_axis_color,
-            self.XAxis: self._x_axis_color,
-            self.YAxis: self._y_axis_color,
-            self.ZAxis: self._z_axis_color,
-            self.AllAxis: self._all_axis_color
-        }
+        if theme:
```
bradymadams, line 53: What happens if theme is None? Is it safe to not set these attributes?

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -238,8 +238,14 @@ def _saveState(self, writing_workspace=False):
             self._storage.setEntryToStore(plugin_id=self.metadata.id, key="selectedResult", data=None)
 
     # Acquires all of the smart slice data from Cura storage and updates the UI
-    def _getState(self, filename=None):
-        all_data = self._storage.getPluginMetadata(self.metadata.id)
+    def _getState(self, filename=None, testFile=False):
```
bradymadams, line 241: Is `filename` used? Seems redundant with `testFile`.

SmartSlicePlugin/tests/SmartSliceTestCase.py
```diff
@@ -1,23 +1,64 @@
...
+from cura.Scene.SliceableObjectDecorator import SliceableObjectDecorator
+
+# TODO have stl/3mf management load files into these variables
+STLFILES = [
+    '/home/shasta/Downloads/block_20x20x40.stl'
+]
+
+DATA = [
+    '/home/shasta/Downloads/plugin_metadata.json'
+]
```
bradymadams, line 17: Let's remove temporary full paths since this will only work on your machine. If it's just temporary and you can replace with relative paths that would be fine for now.

I think we need a better system then just some raw lists. Let's put some more thought into how to manage assets. It'll be imperative that it can pull assets locally (for unit test development) and from the long term storage. It would be much better to couple related assets into object definitions and provide a cleaner way to refer to them than hard-coded indices.

SmartSlicePlugin/tests/SmartSliceTestCase.py
```diff
@@ -1,23 +1,64 @@
...
-        from cura.CuraApplication import CuraApplication
-        from cura.Settings.SettingOverrideDecorator import SettingOverrideDecorator
-        from cura.Scene.SliceableObjectDecorator import SliceableObjectDecorator
+    def loadValidated3mf(self, extension) -> None:
+        # load in stl file
+        self.app.readLocalFile(QUrl.fromLocalFile(STLFILES[0]))
+        # initialize smart slice state
+        extension._getState()
+        # load in validation data
+        extension._getState(testFile=DATA[0])
```
bradymadams, line 47: Why does `_getState` need to be called twice in a row? I would think about if there's a better way to handle this.

bradymadams (changes requested):
`loadValidated3mf` and `loadOptimized3mf` are identical as of now. Is there any reason to have separate functions?

bradymadams:
> I wasn't able to bypass the dialogs that popup when opening a 3mf so instead I implemented a system where an stl is loaded in and then the corresponding smart slice data is loaded in afterwards. Any changes outside the test folder facilitate that system.

I've noticed that in the past too and unfortunately I don't think there's a way to bypass that dialog. One option would be to extract and load the information from the 3MFs ourselves.

> Also I was getting blocked trying to implement a changePrinter function. Looking through relevant code in Cura and UM it seemed that any class referencing a printer was looking for a printer on the network. @bradymadams If you could point me in the direction of where you think the actions are to add a non-networked printer and then change the printer, I should get that done in no time.

I've never looked into this so I don't know off the top of my head where that API is.

SmartSlicePlugin/select_tool/LoadToolHandle.py
colman-hartinger: Ya it's safe to not set these attributes as they are all GUI related, so theme will always be None when running in headless mode

SmartSlicePlugin/SmartSliceExtension.py
colman-hartinger: filename gets passed in when the signal  connects to _getState, so I thought it would be better to pass in the test file separately and load in the SmartSlice data that way

SmartSlicePlugin/select_tool/LoadToolHandle.py
bradymadams: What happens if a unit test calls some code that references one of those attributes?
Do you think a default theme or some default colors should be set instead to help the future reader of the code see what's going on? If I read that code it wouldn't be obvious to me that "oh this is because unit tests run headless". I'd think it was a bug waiting to happen and would want to fix it.

SmartSlicePlugin/tests/SmartSliceTestCase.py
colman-hartinger: I didn't put too much thought into how the files were retrieved since I thought that would get changed around a lot with #465 I'll implement a better system for that

SmartSlicePlugin/SmartSliceExtension.py
bradymadams: Got it. As there's really no re-use of functionality here I think a separate function for loading the plugin metadata from a file would be better. As it is, unless I know about this alternative use by the unit tests this would be confusing to me if I came across this.

SmartSlicePlugin/tests/SmartSliceTestCase.py
colman-hartinger: I wasn't 100% sure about this one but there were a lot of errors when `_getState` was called to set the smartslice data and I noticed that in the regular smartslice code `_getState` was always called first without any objects before being called again with the smartslice data so I just mimicked that.

bradymadams:
@colman-hartinger can you comment on the decision to put the functionality in `SmartSliceTestCase`? To me, managing the state, seems like a separate responsibility. I think it would be best suited in a separate module and/or class.

colman-hartinger:
> I've noticed that in the past too and unfortunately I don't think there's a way to bypass that dialog. One option would be to extract and load the information from the 3MFs ourselves.

That's already what loadValidated3mf is doing, it loads in the stl and then loads in the plugin_metadata.json calling `_getState`. Do you think there should be some function that extracts plugin_metafata.json from a 3mf?


> `loadValidated3mf` and `loadOptimized3mf` are identical as of now. Is there any reason to have separate functions?

Ya that should be consolidated I Just kept them separate because I wasn't exactly sure how #465 was going to handle retrieving models

SmartSlicePlugin/SmartSliceExtension.py
colman-hartinger: wouldn't that result in a lot of code duplication though?

bradymadams:
> That's already what loadValidated3mf is doing, it loads in the stl and then loads in the plugin_metadata.json calling `_getState`. Do you think there should be some function that extracts plugin_metafata.json from a 3mf?

Whatever we think will be easier to use and maintain.

colman-hartinger:
> @colman-hartinger can you comment on the decision to put the functionality in `SmartSliceTestCase`? To me, managing the state, seems like a separate responsibility. I think it would be best suited in a separate module and/or class.

Ya I added the functionality to `SmartSliceTestCase` because the other tests were already inheriting that class So adding the functionality there made sense to me so that different tests aren't inheriting different objects. It can easily be separated out into a separate class if you think that makes more sense.

SmartSlicePlugin/SmartSliceExtension.py
bradymadams: Yes, I was just looking at the diff and didn't notice the rest of the function. Duplicating the function wouldn't be good, but there are alternatives. I'd like for you to figure out why this function needs to be called twice in the first place and then it might become more obvious how to setup the state from an external file.

bradymadams:
> > @colman-hartinger can you comment on the decision to put the functionality in `SmartSliceTestCase`? To me, managing the state, seems like a separate responsibility. I think it would be best suited in a separate module and/or class.
> 
> Ya I added the functionality to `SmartSliceTestCase` because the other tests were already inheriting that class So adding the functionality there made sense to me so that different tests aren't inheriting different objects. It can easily be separated out into a separate class if you think that makes more sense.

I don't think a separate class needs to be inherited. I would use composition here. Put it in a separate class and SmartSliceTestCase can create an instance of it. I think the managing the state of Cura and SmartSlice will get complicated enough that putting it in SmartSliceTestCase is going to make things messy to maintain in the future.

SmartSlicePlugin/select_tool/LoadToolHandle.py
colman-hartinger: Ya a default theme seems like a much better Idea than simply not setting the attributes. I'll implement that

SmartSlicePlugin/tests/SmartSliceTestCase.py
```diff
@@ -1,23 +1,64 @@
...
-    @classmethod
-    def tearDownClass(cls):
+    def setSmartSliceStage(self) -> None:
+        self.controller.setActiveStage("SmartSlicePlugin")
+
+    def clearBuildPlate(self) -> None:
+        self.root.removeAllChildren()
+
+    def setPrintProperty(self, key, value):
+        extruder = self.app.getMachineManager().activeMachine.extruderList[0]
```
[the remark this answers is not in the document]
colman-hartinger: good catch that should be switched out to use the getNodeActiveExtruder function defined in __init__.py

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -261,39 +292,9 @@ def _getState(self, filename=None):
...
-            self.cloud.propertyHandler.resetProperties()
-            self.cloud.updateSliceWidget()
-            self.proxy.updateColorUI()
-
-            self._storage.getPluginMetadata(self.metadata.id).clear()
-
         if job:
             self.cloud.propertyHandler.pauseTracking("extension_updatePropertiesFromJob")
-            self.proxy.updatePropertiesFromJob(job, afterSmartSliceNodeInit)
+            self.proxy.updatePropertiesFromJob(job, self.afterSmartSliceNodeInit(status, results, job, selected_row))
```
bradymadams, line 297: This looks fundamentally different from how it was. Won't this execute the functionality in `afterSmartSliceNodeInit` immediately rather than sending it as a callback?

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,128 @@
...
+    def setSTL(self, file):
+        self.stl = file
+
+    def setThreeMF(self, file):
+        self.threeMF = file
+
+    def loadStlFile(self) -> None:
+        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+
+    def load3mf(self):
```
bradymadams, line 36: Check for consistency in naming. `load3mf` vs `setThreeMF`.

bradymadams (changes requested):
Before I dive in to a deep review it looks like you renamed `SmartSliceTestCase` to `SmartSliceTestStage`. That doesn't seem to address my concern about separating out the functionality of the state management away from the test cases and into it's own class.

We have a 1:1 in a half hour so we can chat about this then.

colman-hartinger:
@bradymadams actually hold off for a second on the review I forgot to change some of the names we were talking about

colman-hartinger:
@bradymadams All right everything should be good to review now. Also I separated out the test case and stage using composition. I think it was causing me problems before since I was trying to build everything from the ground up but now everything is working nicely

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -633,9 +633,10 @@ def createSteps(self, transform_bcs) -> pywim.WimList:
         return steps
 
     def setOrigin(self):
-        controller = Application.getInstance().getController()
-        camTool = controller.getCameraTool()
-        camTool.setOrigin(self.getParent().getBoundingBox().center)
+        if not Application.getInstance().getIsHeadLess():
+            controller = Application.getInstance().getController()
+            camTool = controller.getCameraTool()
+            camTool.setOrigin(self.getParent().getBoundingBox().center)
```
colman-hartinger, line 639: @bradymadams I wasn't really looking setting a mocked camTool. If you think this could cause some problems down the road I'll look into mocking the camTool object

colman-hartinger: it also looks like cura doesn't set any default values when running in headless mode https://github.com/Ultimaker/Cura/blob/master/cura/CuraApplication.py#L902

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,70 @@
...
+    def loadStlFile(self) -> None:
+        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+
+    def loadThreeMF(self):
+        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+
+        with open(self.threeMF, 'r') as data:
+            self.threeMFData = json.load(data)
+
+        self.loadSmartSliceData()
```
AijunHall, line 37: Could you explain if these methods are intended to load the .stl/.3mf from a s3 key/local file path, or from the actual binary data? So in other words, what is the relationship between `file`, `self.stl`, `self.threeMF`, `self.threeMFData`

My understanding of #465 is that much of this file system functionality is supposed to be contained within that issue, within its own class/module. Could you pitch in on this @bradymadams ?

What I have been writing up is intended to have methods that return the raw binary data `tmf` of .3mf/.stl files (from the local system or s3). If these are just example or placeholder methods until 465 then thats fine since I'm sure you will adapt them appropriately. I just want to ask about your intentions with this class, since it also interacts with the Cura API storage/loading that I'm not aware of.

Do we need 2 "file systems" where 1 loads the binary data, and another sets it up for the Cura API loads? Or should it be combined?

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,70 @@
...
+        self.stl = file
+
+    def setThreeMF(self, file):
+        self.threeMF = file
+
+    def loadStlFile(self) -> None:
+        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+
+    def loadThreeMF(self):
+        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
```
AijunHall, line 32: Is there a reason why the `self.stl` is being loaded within this 3mf load method?

colman-hartinger: Ya Cura throws a couple dialogs when you load in 3mf that aren't easy to bypass so this loads in a stl then loads the corresponding `plugin_metadata.json` to set the smartslice data as an alternative.

So for what you're working on you just need some method that retrieves the binary stl and the meta_data,json files

SmartSlicePlugin/tests/SmartSliceTestStage.py
colman-hartinger: `file` is just an argument, `self.stl` is the path to the stl, `self.threeMF` is the path to the plugin_metadata.json file, `self.threeMFData` is a dictionary of the smartslice data.

So right now this class is expecting the local file path, although I expect a lot to change here with the addition of your work. With a bit of tweaking this can be modified to accept a binary stl so I think that's good for now. If you want to stick with loading a 3mf from s3 then there would need to be some functionality to extract plugin_metadata.json from that file. Although most of the 3mf data is irrelevant so you may want to switch to just pulling the metadata.json file from s3

colman-hartinger: and I think 2 filesystems would be redundant, I would just think of this class as a placeholder until #465 is done

SmartSlicePlugin/stage/SmartSliceScene.py
bradymadams: I can't think of a reason to at this point, so how you have it should be fine.

SmartSlicePlugin/tests/SmartSliceTestCase.py
```diff
@@ -31,4 +34,32 @@ def createPrintableModel():
...
+    def tearDown(self) -> None:
+        self.stage.clearBuildPlate()
+
+    def test_anchor_face(self):
+        from SmartSlicePlugin.utils import getPrintableNodes
+        testNode = getPrintableNodes()[0]
+        child = testNode.getChildren()[0]
+        anchor = child.getChildren()[0]
+        face = anchor._surface_type
+        assert face.name == 'Flat'
```
bradymadams, line 61: Why are you using `assert` rather than the assert functions built into `unittest`?

SmartSlicePlugin/tests/SmartSliceTestCase.py
```diff
@@ -31,4 +34,32 @@ def createPrintableModel():
...
-        node.addDecorator(SettingOverrideDecorator())
\ No newline at end of file
+        node.addDecorator(SettingOverrideDecorator())
+
+
+'''
+Example class implementing new state management system
+should be moved or deleted once more tests are written
+executes simple test asserting that the anchor face type on the validated part is flat
+'''
```
bradymadams, line 44: [Doc strings should go under the class.](https://google.github.io/styleguide/pyguide.html#384-classes)

SmartSlicePlugin/tests/SmartSliceTestCase.py
```diff
@@ -1,23 +1,26 @@
...
+from cura.Scene.SliceableObjectDecorator import SliceableObjectDecorator
+from SmartSliceTestStage import SmartSliceTestStage, FileSystem
 
-class _SmartSliceTestCase(unittest.TestCase):
-    @classmethod
-    def setUpClass(cls):
-        cls.app = CuraApplication.getInstance()
+# TODO add path to files in s3 here
+STL = '/block_20x20x40.stl'
+THREEMF = '/plugin_metadata.json'
```
bradymadams, line 14: Seems like there should be a better way to do this than setting module level variables. If you think this will get cleaned up in the model loading in #465 then we can leave it for now.

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,70 @@
...
+        self.extension = SmartSliceExtension()
+        self.stl = None
+        self.threeMF = None
+        self.threeMFData = None
+
+    def setSTL(self, file):
+        self.stl = file
+
+    def setThreeMF(self, file):
+        self.threeMF = file
```
bradymadams, line 26: Why not pass these into the constructor?

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def loadStlFile(self) -> None:
+        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+
+    def loadThreeMF(self):
+        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+
+        with open(self.threeMF, 'r') as data:
+            self.threeMFData = json.load(data)
```
bradymadams, line 35: The file you're reading isn't actually a 3MF, so the naming is misleading.

Setting this attribute, only to call the only function that refers to it isn't great design. It introduces more unnecessary state to the class that needs to be managed. A cleaner design would be:

```
   threeMFData = json.load(data)
   self.loadSmartSliceData(threeMFData)
```

Limiting the scope of variables as much as possible will help result in cleaner and less bug-prone code.

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,70 @@
...
+    def loadSmartSliceData(self):
+        job_dict = self.threeMFData.get("job", None)
+        status = self.threeMFData.get("status", None)
+        results_dict = self.threeMFData.get("results", None)
+        row = self.threeMFData.get("selectedResult", None)  # The row is stored as the order of the results
+
+        self.storage.setEntryToStore('SmartSlicePlugin', 'job', job_dict)
+        self.storage.setEntryToStore('SmartSlicePlugin', 'status', status)
+        self.storage.setEntryToStore('SmartSlicePlugin', 'results', results_dict)
+        self.storage.setEntryToStore('SmartSlicePlugin', 'row', row)
```
bradymadams, line 48: Double quote strings.

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.extension.cloud._onEngineCreated()
+        self.extension._getState()
+
+
+class SmartSliceTestStage:
```
bradymadams, line 54: I think the name of this class is a little misleading. It's not just representing the SmartSlice stage, right? What is it's actual purpose? Answering that question might help you name it better.

SmartSlicePlugin/tests/SmartSliceTestStage.py
```diff
@@ -0,0 +1,70 @@
...
+        self.app = CuraApplication.getInstance()
+        self.controller = self.app.getController()
+        self.scene = self.controller.getScene()
+        self.root = self.scene.getRoot()
+
+    def setSmartSliceStage(self) -> None:
+        self.controller.setActiveStage("SmartSlicePlugin")
+
+    def clearBuildPlate(self) -> None:
+        self.root.removeAllChildren()
```
bradymadams, line 65: I think this will also remove the nodes used by Cura to render the build volume. I'm not sure if that will have an adverse effect on other tests or not, but it might be worth making a note of in case it does come up as a problem in the future.

bradymadams (changes requested):

SmartSlicePlugin/tests/SmartSliceTestCase.py
colman-hartinger: Yup I think this will change with #465

SmartSlicePlugin/tests/SmartSliceTestCase.py
colman-hartinger: I wrote these tests pretty fast as an example to help Aijun if I was going to move to a different project. I'll clean this class up but as these tests stand I don't think they hold too much merit

SmartSlicePlugin/tests/SmartSliceTestStage.py
colman-hartinger: I thought it would be useful to have the ability to set the different file names and since the tests class are inheriting SmartSliceTestCase and not creating a new instance of it for every test this seemed more robust

SmartSlicePlugin/tests/SmartSliceTestStage.py
colman-hartinger: Ya I'll switch it over to just delete the printable, I forgot the root node contains the build plate and a couple other non printing nodes

SmartSlicePlugin/tests/SmartSliceTestStage.py
bradymadams: I'd argue that if you need to change the internal state of a FileSystem instance over and over then it should be re-created over and over. Classes are easiest to use (and less prone to bugs) when they are not constantly mutating state.

If you are setting up the same files for every unittest in a TestCase:

```python
def setUp(self):
  self.files = FileSystem(stl=stl_identifier)
```

bradymadams:
Would FileSystem be better suited as part of the state manager class (currently named SmartSliceTestStage)?

SmartSlicePlugin/tests/SmartSliceTestStage.py
colman-hartinger: Okay I'll change to setting it in the constructor

colman-hartinger:
> Would FileSystem be better suited as part of the state manager class (currently named SmartSliceTestStage)?

Ya I think it would. I separated them out SmartSliceTestStage was a part of SmartSliceTestCase but now it seems like it would be better if they were mergerd

bradymadams (approved):
This looks cleaner. I think we can move forward with this, but let's not be afraid to tweak and update as necessary.

One thing that jumps out at me is the usage of `SmartSliceExtension` and the calls to the private methods. It's definitely not clear, by just reading those calls, why it's necessary, and it's  a little bit of a code smell to me.

colman-hartinger:
> This looks cleaner. I think we can move forward with this, but let's not be afraid to tweak and update as necessary.
> 
> One thing that jumps out at me is the usage of `SmartSliceExtension` and the calls to the private methods. It's definitely not clear, by just reading those calls, why it's necessary, and it's a little bit of a code smell to me.

Ya I was trying to get rid of accessing private members by emiting the signals that trigger those functions but the `engineCreated` signal caused more problems than necessary when running in headless mode

AijunHall (approved):
Looks clearer to me as well since `CuraStateManagement` is separated out.

PR #476, opened 2021-03-05
Unittest Model Retrieval

AijunHall:
This is working off the basis that we are keeping the S3 bucket private.

Currently `CuraStateManagement`'s loading of Parts works from a class attribute that is a single model, but I've included methods that group models together in lists. I thought it would be beneficial to have the option of being able to pull from lists of models, as well as being able to load single models for tests

This also sets up a local directory for downloading the S3 objects. Ideally large S3 folders would be downloaded as they are used, instead of all at once as they are here. That might require separating out acquiring the S3 paths vs. actually downloading the files. But I'd like to hear some thoughts about that, in relation to our new management system

I would also like to know if there's any other 'filesystem' functionality that needs to be added in or changed here to accommodate Cura, or SmartSlice

Also if/how I would need to add any of this to the github actions (without pushing boto3 credentials of course)


SmartSlicePlugin/tests/FileLoader.py
```diff
@@ -0,0 +1,171 @@
...
+    Local system functionality
+    '''
+    def getModel(self, local_file_path: str) -> bytes:
+        if not os.path.isfile(local_file_path):
+            return None
+
+        with open(local_file_path, 'rb') as f:
+            tmf = f.read()
+
+        return tmf
```
colman-hartinger, line 48: Normally in Cura you'll load in a part using `app.readLocalFile(QUrl.fromLocalFile(somefile.stl))` can you load in binary data the same way?

SmartSlicePlugin/tests/ExampleTests.py
```diff
@@ -13,6 +9,13 @@ class ExampleTests(unittest.TestCase):
     '''
     @classmethod
     def setUpClass(cls) -> None:
+        # Loading up STL and 3MF/smartSliceData proposed method/example
+
+        cls.fileloader = _S3FileLoader('s3://is-cura-ui-test-bucket')
+
+        STL = cls.fileloader.getModel('stl_folder/model.stl')
+        THREEMF = cls.fileloader.getMetaDataJson('3mf_folder/plugin_metadata.json')
```
colman-hartinger, line 17: This is my fault for leaving the THREEMF object but we should probably get rid of any mention of 3mfs since what is actually being loaded is an stl and a json file

colman-hartinger (changes requested):
This is just an initial code review, I haven't tested the code yet. That being said everything is looking pretty good.

In  `LocalFileLoader` and  `S3FileLoader` there is a lot of code duplication that can be consolidated. 

For Example LocalFileLoader.getModel and LocalFileLoader.getMetaDataJson perform essentially the same task and have almost identical code. A cleaner solution could be something like this.

```
class LocalFileSystem:
    def __init__(self, stl=None, pluginMetaData=None):
        self.stl = stl
        self.pluginMetaData = pluginMetaData

    def getModel(self, local_file_path: str):
        if not os.path.isfile(local_file_path):
            return None

        if self.stl:
            return self.loadFile()

        if self.pluginMetaData:
            return self.loadJson() 
```

Also see my other comment about getting rid of using `3mf` it doesn't look like you are extracting the plugin_metadata.json from a 3mf so lets name that accordingly.

SmartSlicePlugin/tests/FileLoader.py
```diff
@@ -0,0 +1,171 @@
...
+        '''
+        Return plugin_metadata.json
+        '''
+        raise NotImplementedError()
+
+    def getMetaDataJsons(self, folder_path: str) -> Tuple[Dict]:
+        '''
+        Return all plugin_metadata.json files in a folder, as a tuple
+        '''
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
AijunHall: `_FileLoader` is just a base class that outlines the file retrieving methods. I separate out the local system and s3 functionality classes from here since both require different weird path manipulations, but have the same purpose.

If you think it would be cleaner to combine these into a single class with both local, and s3 accessibility, I can do that.

SmartSlicePlugin/tests/FileLoader.py
AijunHall: This is what I was mainly confused about in how this would interact with Cura. Now its more clear that the part should be loaded in that way.

This only uses the string path so I'll change that.
https://doc.qt.io/qtforpython/PySide6/QtCore/QUrl.html#PySide6.QtCore.PySide6.QtCore.QUrl.fromLocalFile

bradymadams (changes requested):
I'll wait for you and @colman-hartinger to iterate a little bit before I provide a deeper review, but I have a few comments:

1. The configuration of the file loader should be specified externally (local vs s3 and directory names).
2. Some of the naming is a little risky (getModel vs getModels), especially since the argument lists are the same. It would be really easy to type in the wrong one and then be confused why it's not working. I'd recommend using `getAllModels` or `listModels` to differentiate more. Likewise for other functions. "List" is the more common verb in this context, in my experience.
3. Are the list functions needed?
4. You guys might think about how to robustly store the files so they're flexible for minor changes. For instance iterating on the '.json' extension could be a problem if we ever need to add another JSON file. I'd probably consider explicit file names and organizing jobs by folder.
5. We might need to be careful about indiscriminately downloading all the files from S3 everytime one a list function is called. That could get time consuming (and also increase AWS bills).
6. Underscore prefixes on names indicate they are "private". A private class usually means it's only intended to be used internally within that file/module. I don't think that's the case for the file loader classes.

SmartSlicePlugin/tests/FileLoader.py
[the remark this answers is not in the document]
bradymadams: The base abstract class makes sense and should be kept. It helps with typing, consistency, and consolidating common code.

```python
def foo(file_loader: FileLoader):
  # do something with file_loader
```

`foo` doesn't need to know what exact type `file_loader` is, just that it's something that has implemented the `FileLoader` methods. Very common OOP practice.

AijunHall:
Since moving away from boto3 makes this more straightforward, It seemed most appropriate to me to use what is already in `CuraStateManagement` for model loading.

SmartSlicePlugin/tests/CuraStateManagement.py
```diff
@@ -29,11 +30,13 @@ def changePrinter(self, machine):
         machine_manager = self.app.getMachineManager()
         machine_manager.addMachine(machine)
 
-    def loadPart(self) -> None:
-        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+    def setSTL(self, stl_path) -> None:
+        if os.path.isfile(stl_path) and stl_path.lower().endswith('.stl'):
+            self.stl = stl_path
```
colman-hartinger, line 35: As @bradymadams was noting in my PR for the state management system, it's better to just create a new object instead of mutating the current object so lets add this logic to the constructor

SmartSlicePlugin/tests/CuraStateManagement.py
```diff
@@ -29,11 +30,13 @@ def changePrinter(self, machine):
...
 
-        if self.smartSliceData:
-            self.loadSmartSliceData()
+    def setSmartSliceData(self, plugin_metadata_path) -> None:
+        if os.path.isfile(plugin_metadata_path) and plugin_metadata_path.lower().endswith('json'):
+            self.smartSliceData = plugin_metadata_path
```
colman-hartinger, line 39: same note as above

SmartSlicePlugin/tests/README.md
```diff
@@ -28,10 +28,13 @@ class ExampleTests(_SmartSliceTestCase):
...
     - `clearBuildPlate`
     - `setPrintProperty`
     - `loadValidated3mf`
     - `loadOptimized3mf`
-    - `loadStlFile`
\ No newline at end of file
+    - `loadStlFile`
+
+### Loading Models
+- Models are loaded in as .stl files or plugin_metadata.json files
```
colman-hartinger, line 40: Just to clear up any confusion here models are loaded in as .stl files smartslice data is loaded in with a correspoinding plugin_metadata,json file

colman-hartinger (changes requested):
This doesn't look like any new functionality has been added compared to whats already in place. Lets create the repo for storing stl and json data and then build a FileSystem class for loading that data. The FileSystem class can be added to the CuraStateManagement class using composition.

Also as noted in some of my comments we should we should set the stl and plugin_metadata filepath in the constructor of the class. This is because it makes more sense to just have a single model for each test case and having a test case that calls multiple different models could be hard to maintain and troubleshoot if/when something goes wrong.

SmartSlicePlugin/tests/README.md
```diff
@@ -28,10 +28,16 @@ class ExampleTests(_SmartSliceTestCase):
...
     - `loadOptimized3mf`
-    - `loadStlFile`
\ No newline at end of file
+    - `loadStlFile`
+
+### Loading Models
+- Models are loaded in as .stl files
+
+### Loading SmartSliceData
+- SmartSlice data is loaded in with a corresponding plugin_metadata.json file
```
bradymadams, line 43: A little info on _how_ to load models and SmartSlice data would be useful.

bradymadams (changes requested):
1. There should be a configuration for the root directory of the model files, in my opinion.
2. I think it would be most robust to have a single function to load a model by relative path. The code that finds the files can determine what files are in there (STL only or STL+JSON) and take the appropriate action.

AijunHall:
will start new branch/PR because of change in ticket goal

PR #477, opened 2021-03-07
R21.0 welcome screen

dougkenik:
I still need to get the complete URL for the welcome video, but a draft review would be great here:

- Created a welcome screen on login
- Added videos for welcome and tutorials on a second screen after welcome
- Video screen is accessible via Extensions or the left toolbar (?)
- The videos are GIFS I created and hosted on our website, so we can easily update them
- Clicking on a video takes the user to the Help webpage instead of displaying the video - displaying the video in-product was going to be too much work
- I decided not to have a "Don't show me this again" checkbox on the Welcome screen in case we add more functionality and want to re-purpose. It's very easy to dismiss it right now
- I added a Preferences class which can be used around our SmartSlice application. We had preferences strewn about, so I wanted to consolidate

We could add the videos in-product if we created them as GIF's. I am unsure how useful this would be, but I am open to the discussion.


SmartSlicePlugin/stage/ui/SmartSliceWelcome.qml
```diff
@@ -0,0 +1,432 @@
...
+                        anchors {
+                            top: parent.top
+                            left: parent.left
+                            leftMargin: 2 * UM.Theme.getSize("wide_margin").width
+                            rightMargin: 2 * UM.Theme.getSize("wide_margin").width
+                        }
+
+                        width: parent.width - anchors.leftMargin - anchors.rightMargin
+
+                        source: 'https://www.tetonsim.com/hubfs/tutorial2.gif'
```
bradymadams, line 313: 1. How stable do we think these links are? Is there a risk that someone might change this without knowing the plugin is referencing it?
2. Do we have a consolidated place in the plug-in code (outside of QML) where we are specifying links. If they're scattered, it might be a good idea to consolidate so when we need to change something we can easily find all the links and change the necessary ones without risk of missing any.

bradymadams (changes requested):
This is just a brief review of the functionality and a couple other comments. I'll let @colman-hartinger review the code first.

1. Unit tests are failing.
2. I don't see any gifs, and I see these errors:

```
QML AnimatedImage: Error Reading Animated Image File https://www.tetonsim.com/hubfs/tutorial1.gif
```

image

colman-hartinger:
It looks like the unit tests are failing because of one of the mocked objects not being initialized right. I'll look into that but I'm not sure that it was a problem introduced by this PR. I haven't gone through the code yet but when you aren't in full screen the new welcome screen gets a little funky. I think that it should look good for all allowable screen sizes in cura

welcome

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -256,10 +256,10 @@ def __init__(self, connector):
         self._number_of_timeouts = 20
         self._timeout_sleep = 3
 
-        self._username_preference = "smartslice/username"
-        self._app_preferences = Application.getInstance().getPreferences()
+        self._preferences = connector.extension.preferences
```
colman-hartinger, line 259: This is the line that is causing the unit tests to fail. I don't see anything wrong with calling the preferences from the cloud connector so I'll add a fix so that object is properly mocked

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -256,10 +256,10 @@ def __init__(self, connector):
...
 
         #Login properties
+        self._first_name = "Friend"
```
colman-hartinger, line 262: Is the welcome screen supposed to display the users name? After I login it still says "Welcom to Smart Slice, Friend"

SmartSlicePlugin/SmartSliceCloudConnector.py
```diff
@@ -785,7 +795,7 @@ def _onEngineCreated(self):
 
         Application.getInstance().getMachineManager().printerConnectedStatusChanged.connect(self._refreshMachine)
 
-        if self.app_preferences.getValue(self.debug_save_smartslice_package_preference):
+        if self.app_preferences.getPreference(Preferences.SaveDebug):
```
colman-hartinger, line 798: @bradymadams should we get rid of the debug package? It seems obsolete now that we can just save 3mfs with smartslice data

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -181,6 +193,24 @@ def accountUrl(self):
     def tooltipLocations(self):
         return {"left" : 1, "right" : 2, "top" : 3, "bottom" : 4}
 
+    @pyqtSlot(bool)
+    def setIntroScreenVisible(self, intro_screen_visibile: bool):
```
colman-hartinger, line 197: I think this should be renamed to `setIntoScreenVisibility` since this is also used to turn off the visibility

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -181,6 +193,24 @@ def accountUrl(self):
...
+        self._intro_screen_visible = intro_screen_visibile
+        self.introScreenVisibleChanged.emit()
+
+    @pyqtProperty(bool, notify=introScreenVisibleChanged)
+    def introScreenVisible(self):
+        return self._intro_screen_visible
+
+    @pyqtSlot(bool)
+    def setWelcomeScreenVisible(self, welcome_screen_visibile: bool):
```
colman-hartinger, line 206: same issue as above comment

colman-hartinger (changes requested):

SmartSlicePlugin/SmartSliceCloudConnector.py
bradymadams: The "debug" package is a slightly different format than what is saved by Cura. The reason is because we developed our 3MF format for the back-end before Cura implemented the plugin metadata save logic and they picked a different format. Eventually, I'd like to have the back-end recognize both formats (or move to all JSON). But for now I think there's a benefit to leaving it in. It allows Jeff and others to quickly save a 3MF that they can submit to the smartslice engine locally for dev and testing.

SmartSlicePlugin/SmartSliceCloudConnector.py
dougkenik: Is should display user name. "Firend" is a default. I'll check on this, but it should be displaying your name

SmartSlicePlugin/stage/ui/SmartSliceWelcome.qml
dougkenik: 1. They are on a landing page, which we can change at any time. They can change at any time, which was the reason for putting them on the landing page. For example, when this functionality is released, the welcome video needs to change with he UI. If there is a better approach, I am open to it - I thought this was best so we can update the help videos and these GIFS all in Hubspot
2. No - good idea, and I will look into that. Should this be a different ticket though?

bradymadams: 1. I wasn't suggesting they not be editable or that they be put in the repo. I agree with hosting them so we can update them. What I don't have a feel for is how hubspot hosts them and who has access to them. Is it possible for someone to come across the gif, think they're cleaning something up, and unintentionally rename or move it? If you think that's unlikely, it's fine - it's just something I always think about when I see a URL hardcoded client side.
2. If there's not a central place for URLs yet, then yes, separate ticket.

bradymadams: And the more serious problem would be if we ever moved away from hubspot - all those links would get broken.

dougkenik: > 1. I wasn't suggesting they not be editable or that they be put in the repo. I agree with hosting them so we can update them. What I don't have a feel for is how hubspot hosts them and who has access to them. Is it possible for someone to come across the gif, think they're cleaning something up, and unintentionally break the link by renaming or moving it? If you think that's unlikely, it's fine - it's just something I always think about when I see a URL hardcoded client side.
> 2. If there's not a central place for URLs yet, then yes, separate ticket.

If we want to host them somewhere else besides in hubspot, then we would need to agree as to where. If you have an idea of how to do it, or how you would prefer it done - then please let me know.

dougkenik: There is definitely a chance that the videos could get deleted or replaced on the landing page - or the landing page could get deleted entirely. The only real steps we can take to protect it is to name it in such a way that it won't get deleted.

bradymadams: Okay :+1:

dougkenik:
@colman-hartinger - screen size is dynamic now

SmartSlicePlugin/SmartSliceCloudProxy.py
```diff
@@ -134,6 +139,9 @@ def __init__(self, metadata: "SmartSliceExtension.PluginMetaData") -> None:
 
     # Properties (mainly) for the sliceinfo widget
 
+    introScreenVisibleChanged = pyqtSignal()
+    welcomeScreenVisibleChanged = pyqtSignal()
```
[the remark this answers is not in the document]
dougkenik: Intro screen contains the welcome message. Welcome screen contains the videos. You can navigate to the welcome screen directly through product (don't want to show the intro screen if someone tries to access the welcome screen in product)

SmartSlicePlugin/components/SmartSliceTooltip.qml
```diff
@@ -91,7 +91,7 @@ UM.PointingRectangle {
             textFormat: Text.RichText
             font: UM.Theme.getFont("default_bold")
             color: UM.Theme.getColor("tooltip_text")
-            renderType: Text.NativeRendering
+            // renderType: Text.NativeRendering
```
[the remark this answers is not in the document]
dougkenik: The tooltip text was displaying very poorly in the video tooltips. It is because of Native Rendering. This has happened before when we implemented the login screen

dougkenik:
> Everything looks a lot better now when not in full screen although the screen with the videos still has some overlapping text. Have you thought about doing this as a popup? I think a popup would be less intrusive and allow the user to switch between messing around with the part in smart slice and choosing which tutorial to watch

Can you share a sreenshot of what the overlaping text looks like? We agreed on the current design as-is. We could allow the user to drag and resize the welcome screen, as well as access the scene behind it if necessary, but when you click on a vide, it opens the web browser, which allows you to dismiss the welcome screen and follow along. Is that not good enough? What features of a popup are better?

colman-hartinger:
Screenshot 2021-03-10 11:51:21

heres the welcome screen when I'm not in using cura in fullscreen.

>What features of a popup are better?

The only benefits I could see for a popup over what is currently in place is that it would allow the user to go through a couple different workflow tutorials without having to open and close the welcome screen. I also personally believe that non "mission critical" aspects of a program should be as in-obtrusive as possible. That being said I don't think there's that big of a difference between 2 or 3 clicks to remove the intro/welcome screen vs 1 click to hide a popup

dougkenik:
The welcome video is now in help, so this can turn into a full-fledged review

colman-hartinger (approved):
Everything looks good to me. Also I'll add the fix for the failing tests in r21.0-tests since it will be a cleaner  solution to pass the the correct objects in using CuraStateManagement.

SmartSlicePlugin/SmartSlicePreferences.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def setPreference(self, preference: str, value: Any):
+        pref = self._extend(preference)
+        if self._app_preferences.getValue(pref) is None:
+            self._app_preferences.addPreference(pref, value)
+        else:
+            self._app_preferences.setValue(pref, value)
+
+    def _extend(self, preference: str) -> str:
+        t = self._preference_base + preference
```
bradymadams, line 41: Unnecessary line?

bradymadams (commented):
One trivial thing I noticed. Some of the JS in the QML is missing semi-colons.

bradymadams (approved):

PR #482, opened 2021-03-10
R21.0 cache tests

colman-hartinger:
Made some changes so that tests that need part/smart slice data won't run until the part/data has finsihed loading and added some tests to that the cache system is working properly


colman-hartinger:
closing this to cherrypick changes from r21.0-tests not r21.0

Issue #359, opened 2020-11-24
Ensure the arrows for SmartSlice Loads display correctly


jake-kirkland-teton:
Additionally, make sure the rotation tool appears as expected.

Issue #358, opened 2020-11-24
Check SmartSlice Job Anchors


jake-kirkland-teton:
Write test(s) for ensuring anchors are defined and sent correctly. 

Check to make sure that the face selection type is maintained.

Issue #357, opened 2020-11-24
Checking SmartSlice Job Loads


jake-kirkland-teton:
Write unit test(s) for ensuring that that loads for a job are defined and passed to the back end correctly.

bradymadams:
@jake-kirkland-teton I like the idea of differentiating these issues. Let's use the unittest label since it will be more consistent and we can filter on it.

jake-kirkland-teton:
Absolutely. That is an excellent idea.

PR #483, opened 2021-03-10
R21.0 cache system tests

colman-hartinger:
Unit tests for testing SmartSlicePropertyHandler initialization and when properties are changed after a part with smart slice data is loaded in. Also implemented fix for state management so unit tests only run after part/data has been loaded in


colman-hartinger:
Here is the part and plugin_metadata.json I used for local testing

[block_20x20x40.zip](https://github.com/tetonsim/is-cura-ui/files/6119415/block_20x20x40.zip)

SmartSlicePlugin/tests/CuraStateManagement.py
```diff
@@ -20,7 +24,17 @@ def __init__(self, stl=None, smartSliceData=None):
...
 
         self.pluginData = PluginMetaData()
-        self._extension = SmartSliceExtension()
+        self.extension = SmartSliceExtension()
+        self.proxy = SmartSliceCloudProxy(self.pluginData)
+        self.cloud = SmartSliceCloudConnector(self.proxy, self.extension)
+
+        self.extension.smartSliceDataLoaded.connect(self._smartSliceDataLoaded)
+
+        self.smartSliceDataLoaded = False
```
bradymadams, line 33: Setting this as a class level attribute will get buggy once you want to test things like loading a new part in the middle of a SmartSlice session, or something similar where you call `loadAndWait` more than once in a unit test. I'd recommend moving `_smartSliceDataLoaded` and `smartSliceDataLoaded` into a different scope that isn't sustained after the fact. I don't see why you couldn't contain everything within `loadAndWait`.

From a design perspective - this is something I harp on a lot in everyone's PRs - reducing the amount of state you have to manage. The less state you manage in your code, the less opportunity for bugs. When it comes to classes we often get enticed by throwing information into the class, but if it's not necessary, it's just adding more state to manage that could be problematic. It's better to use the lowest level scope possible (if the performance doesn't suffer).

bradymadams (changes requested):
Looks pretty good. I have a few things...

1. I think we should add more tests for various different types of properties. e.g. `ExtruderProperty`,  `GlobalProperty`, `ActiveQualityGroup`, etc.. We should be trying to get as much code coverage as possible.

bradymadams:
Also, it looks like the tests don't require a specific model. It might be best, in these cases, to just use the basic, existing, code that builds the cube model programmatically (not sure what file that's in, but if you don't know what I'm talking about, let me know - I'll search for it).

SmartSlicePlugin/SmartSliceExtension.py
```diff
@@ -241,6 +244,10 @@ def _saveState(self, writing_workspace=False):
     def _getState(self, filename=None):
         all_data = self._storage.getPluginMetadata(self.metadata.id)
 
+        # Initialize property handler for unit tests
+        if self.cloud.propertyHandler is None:
+            self.cloud.propertyHandler = SmartSlicePropertyHandler(self.cloud)
```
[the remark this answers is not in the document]
colman-hartinger: In the regular app `SmartSlicePropertyHandler` gets initialized on the `engineCreated` Signal which never gets emitted when running in headless mode. Thats why before I was calling `SmartSliceCloudConnector._onEngineCreated' but that ended up causing some bugs as well. I was going back and forth on the best possible solution of where to initialize `SmartSlicePropertyHandler` for the purpose of unit testing and ended with this but I'll try setting it from the unit test state management because I agree that the unit tests shouldn't interfere with the main codebase.

colman-hartinger:
> Also, it looks like the tests don't require a specific model. It might be best, in these cases, to just use the basic, existing, code that builds the cube model programmatically (not sure what file that's in, but if you don't know what I'm talking about, let me know - I'll search for it).

We could use the cube from SmartSliceTestCase but we will still need to generate smartslice data for it. The way the system is set up now is that a part is loaded in, the `fileCompleted` signal triggers `SmartSliceExtension._getState` which emits a signal letting the tests know that the smart slice data is loaded as well

bradymadams:
> > Also, it looks like the tests don't require a specific model. It might be best, in these cases, to just use the basic, existing, code that builds the cube model programmatically (not sure what file that's in, but if you don't know what I'm talking about, let me know - I'll search for it).
> 
> We could use the cube from SmartSliceTestCase but we will still need to generate smartslice data for it. The way the system is set up now is that a part is loaded in, the `fileCompleted` signal triggers `SmartSliceExtension._getState` which emits a signal letting the tests know that the smart slice data is loaded as well

Can't we provide some basic, initial, smartslice data? Just as when someone loads an STL initially and goes into the SmartSlice stage? That seems like a common use case for our unit tests - load a model, and then test smartslice from it's initial state.

SmartSlicePlugin/tests/ExampleTests.py
```diff
@@ -11,18 +11,18 @@ class ExampleTests(unittest.TestCase):
     Example class implementing new state management system
     should be moved or deleted once more tests are written
     '''
-    @classmethod
-    def setUpClass(cls) -> None:
-        cls.curaState = CuraStateManagement(stl=STL, smartSliceData=THREEMF)
 
     def setUp(self) -> None:
-        self.curaState.loadPart()
+        self.curaState = CuraStateManagement(stl=STL, smartSliceData=THREEMF)
```
AijunHall, line 16: What's the reasoning for creating a new management system per test, instead of having it  at the class level?

colman-hartinger: this was so you could use curaState.waitAndLoad multiple times within a test case but now that all the logic is contained within a single function this should be moved back to setUpClass

bradymadams: I think it's fine to leave it here if it doesn't hurt performance. However, I don't have a particularly strong argument for the state management object, so I'll let you decide.

But @AijunHall generally, the reason you favor `setUp` over `setUpClass` is to reduce the scope of your objects and have a clean program state before you run your test. It's important that you avoid tests from effecting the state of tests that have yet to be run.

```python
class User:
    def __init__(self, username, password):
        self.username = username
        self.password = password

class MyTests(TestCase):
    @classmethod
    def setUpClass(cls):
        cls.user = User('joe', 'TopS3cr3t')
        
    def test_login(self):
        success = login(user)
        self.assertTrue(success)
        
    def test_bad_password_login(self):
        cls.user.password = 'wrongPassword'
        success = login(user)
        self.assertFalse(success)
```

If `test_bad_password_login` runs before `test_login` then both will fail. If you instead set the user in `setUp` the order of the tests wouldn't be a problem.

colman-hartinger:
> Can't we provide some basic, initial, smartslice data? Just as when someone loads an STL initially and goes into the SmartSlice stage? That seems like a common use case for our unit tests - load a model, and then test smartslice from it's initial state

Even when running in headless, loading a part and switching to the smartslice stage should initialize basic smartslice data just as when running normally.

SmartSlicePlugin/tests/CuraStateManagement.py
```diff
@@ -50,20 +56,68 @@ def loadSmartSliceData(self):
             storage.setEntryToStore(self.pluginData.id, "results", results_dict)
             storage.setEntryToStore(self.pluginData.id, "row", row)
 
-            self._extension.cloud._onEngineCreated()
-            self._extension._getState()
+    def loadAndWait(self):
+        self.waitForLoad = False
```
bradymadams, line 60: Why does this need to be a class attribute?

bradymadams (changes requested):
`TestCache` sort of implies to me it's going to test the cache of all the properties we cache/restore in `SmartSliceProperty.py`. For those we don't have unit tests for yet -- can you comment on which of those you think should fall under this PR and which should be included under a different ticket (or have a ticket created to test them)?

Historically, we've had a lot of buggy behavior when it comes to the property tracking code so having some good code coverage here could be extremely helpful going forward.

colman-hartinger:
> can you comment on which of those you think should fall under this PR and which should be included under a different ticket (or have a ticket created to test them)?

Ya I left a decent amount of properties out since #357 #358 and #359 all  deal with testing the cache for different smart slice properties (although I do think that they should be grouped into one ticket). Given that I think that testing

1.  ToolProperty
2. Transform
3. ActiveExtruder

Should be added to this PR So Ill get on that. I think with the addition of those 3 and 357-359 we will have full coverage of smartSlicePropertyHandler and SmartSliceProperty

SmartSlicePlugin/tests/CuraStateManagement.py
colman-hartinger: when It wasn't a class attribute the signal fire would change the value in a different scope so it would get hung up in the while loop forever

bradymadams: Try using the `nonlocal` keyword. It would be a better solution than setting a class attribute.

https://stackoverflow.com/questions/1261875/python-nonlocal-statement

AijunHall:
I see you've created the new issue for the race condition in TestSmartSliceMesh, but I'm also getting a couple inconsistent unit test results from these new cache system tests. 

A couple of the cache system tests all produce this same traceback from `self.props.cancelChanges()`. Is this related to the same TestSmartSliceMesh race condition or is this a separate thing?

```
File "SmartSlicePlugin/tests\test_cache_system.py", line 76, in test_material_change
    self.props.cancelChanges()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSlicePropertyHandler.py", line 455, in cancelChanges        
    self.restoreCache()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSlicePropertyHandler.py", line 201, in restoreCache
    p.restore()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSliceProperty.py", line 389, in restore
    self._extruder.restore()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSliceProperty.py", line 251, in restore
    self._node.callDecoration("setActiveExtruder", extruder_list[self._active_extruder_index].id)
TypeError: list indices must be integers or slices, not NoneType
```

colman-hartinger:
> I see you've created the new issue for the race condition in TestSmartSliceMesh, but I'm also getting a couple inconsistent unit test results from these new cache system tests when I run the tests over and over.
> 
> A couple of the cache system tests all produce this same traceback from `self.props.cancelChanges()`. Is this related to the same TestSmartSliceMesh race condition or is this a separate thing?

No I think this is a separate condition. I haven't seen that failure come up when I was running tests locally but it seems like there's a separate condition where some of tests run before an active extruder can be initialized. I'll get that fixed

colman-hartinger:
@AijunHall I think I was able to fix the extruder issue you were mentioning by reseting the properties better for each test. would you mind running the changes a couple times and see if you get the error again?

AijunHall:
@colman-hartinger 

I've run the changes for a while now, and I still get it - although definitely less so now. I'll continue to look into this to see if I can give you any more specific information.

Here's the traceback (comes from a different test)

```
======================================================================
ERROR: test_tool_property_change (test_cache_system.TestCache)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "SmartSlicePlugin/tests\test_cache_system.py", line 88, in test_tool_property_change
    self.props.cancelChanges()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSlicePropertyHandler.py", line 455, in cancelChanges        
    self.restoreCache()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSlicePropertyHandler.py", line 201, in restoreCache
    p.restore()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSliceProperty.py", line 389, in restore
    self._extruder.restore()
  File "C:\dev\Cura\plugins\SmartSlicePlugin\SmartSliceProperty.py", line 251, in restore
    self._node.callDecoration("setActiveExtruder", extruder_list[self._active_extruder_index].id)
TypeError: list indices must be integers or slices, not NoneType

----------------------------------------------------------------------
Ran 11 tests in 15.490s

FAILED (errors=1)
```

bradymadams:
Going to close this for now since we're re-focusing dev effort from this. @colman-hartinger when you return let's get the unit test branches cleaned up. If possible, let's get all of the unmerged work into a single branch that we can reference in the future if necessary.

colman-hartinger:
@bradymadams I think if this branch is merged into r21.0-tests then all the unit test work will be clean and consolidated to a single branch

PR #484, opened 2021-03-11
R21.0 unit test filesystem

AijunHall:
Tried to consolidate the model path retrieval, and the plugin metadata retrieval into `getModelData()`

I also moved the json unpacking functionality to the File System as well.

I wanted to get as much of the path strings out of the State Management System as possible, so I thought it might be more appropriate to initialize the File System with a default model to use in unit testing, instead of Cura.

The README also now defines how files should be named and structured to be properly loaded. Let me know your thoughts on this.


AijunHall:
@colman-hartinger The two initial default files in the model storage repo are just the same ones from the zip you showed here:

https://github.com/tetonsim/is-cura-ui/pull/483#issuecomment-796283348

SmartSlicePlugin/tests/CuraStateManagement.py
```diff
@@ -22,36 +20,38 @@ def __init__(self, stl=None, smartSliceData=None):
...
 
     def changePrinter(self, machine):
         machine_manager = self.app.getMachineManager()
         machine_manager.addMachine(machine)
 
-    def loadPart(self) -> None:
-        self.app.readLocalFile(QUrl.fromLocalFile(self.stl))
+    def loadPart(self, model_path=None) -> None:
+        if not model_path:
+            stl_path, smartSliceData = self.filesystem.default_model
```
bradymadams, line 34: Use consistent naming styles. For local variables we use underscore separated lower case.

SmartSlicePlugin/tests/FileSystem.py
```diff
@@ -0,0 +1,45 @@
+from typing import Dict, Tuple
+
+import json
+import os
+
+class FileSystem:
```
bradymadams, line 6: I think `FileSystem` isn't a great name for this class. Something like `ModelLoader`, `ModelRetriever`, etc. might be a better representation of what it's actually doing.

SmartSlicePlugin/tests/FileSystem.py
```diff
@@ -0,0 +1,45 @@
...
+    '''
+    Filesystem for loading models and SmartSlice metadata into unittests
+    '''
+    def __init__(self, root_dir='SmartSlicePlugin/tests/models'):
```
bradymadams, line 10: Hardcoding path separators usually isn't good practice. I'd see if you can avoid that. I'd probably set `root_dir` to default to `None` (`Optional[str]`), and check for None in the body of the constructor - if it's None then build the path relative to `os.getcwd()`.

SmartSlicePlugin/tests/FileSystem.py
```diff
@@ -0,0 +1,45 @@
...
+        self.root_dir = root_dir
+
+        self.default_model = self.getModelData('default.stl')
+
+    def getSmartSliceData(self, metadata_path) -> Tuple[int, Dict]:
+        with open(metadata_path, 'r') as data:
+            threeMFData = json.load(data)
```
bradymadams, line 17: This name is misleading. The data isn't in 3MF format.

bradymadams (changes requested):
1. You're doing a good job of type annotating the returns. Add type annotations to arguments as well.
2. I think it would be more extendible to store models in a directory and refer to a model by it's directory name, rather than by the STL.

For example:

```
root/
  model_A/
     ... some files ..
  model_B/
     ... some files
```

And then request to load a model by the name in the directory.

That will allow you to use specific file names within a directory and you don't have to iterate directories and assume what the file holds based off extension (which could be bug prone for JSON formats).

AijunHall:
> I think it would be more extendible to store models in a directory and refer to a model by it's directory name, rather than by the STL.

Will do. How about something like this in regards to having specific consistent filenames to refer to.

```
root/
  model_A/
     model.stl
     plugin_metadata.json
  model_B/
     model.stl
     plugin_metadata.json
```

I originally thought something like this would be too cumbersome to setup in the first place, so if we go down this path, maybe we could come up with a way to properly setup the models and/or SmartSlice data to the models dir

bradymadams:
> I originally thought this would be too cumbersome to setup in the first place, so if we go down this path, maybe we could come up with a way to properly setup models and/or SmartSlice data to this dir structure

Your setup looks good to me. Can you elaborate on where you think the complications are?

AijunHall:
> Can you elaborate on where you think the complications are?

I don't think there are complications, I was just wondering about how to avoid setting these up manually - depending on how often we would need to add/change models

bradymadams (approved):
Looks good to me. I'll merge upon approval from @colman-hartinger

SmartSlicePlugin/tests/ModelLoader.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, root_dir: Optional[str]=None):
+        if root_dir:
+            self.root_dir = root_dir
+        else:
+            self.root_dir = self.buildDefaultRootDir()
+
+        self.default_model = self.getModelData('default')
+
+    def getSmartSliceData(self, metadata_path: str) -> Tuple[int, Dict]:
```
colman-hartinger, line 26: I would have this return `Tuple[any, any, any, any] since status, job_dict ... can also be None

SmartSlicePlugin/tests/ModelLoader.py
```diff
@@ -0,0 +1,50 @@
...
+            metadata = json.load(f)
+
+        status = metadata.get("status", None)
+        job_dict = metadata.get("job", None)
+        results_dict = metadata.get("results", None)
+        row = metadata.get("selectedResult", None)  # The row is stored as the order of the results
+
+        return status, job_dict, results_dict, row
+
+    def getModelData(self, model_dir_path: str) -> Tuple[str, Tuple[int, Dict]]:
```
colman-hartinger, line 37: I would have the return type here be Tuple[any, any] and then add type annotations for smartslice_data and full_stl_path within the function

colman-hartinger (changes requested):
Everything is looking pretty good and runs as expected. There do seem to be some mislabeled return types in ModelLoader.py once that is fixed I think this should be good to go

SmartSlicePlugin/tests/ModelLoader.py
bradymadams: If the type can be None I think using `Optional[T]` is better than `Any`.

bradymadams: If the type can be None using `Optional[T]` is the standard annotation.

SmartSlicePlugin/tests/ModelLoader.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, root_dir: Optional[str]=None):
+        if root_dir:
+            self.root_dir = root_dir
+        else:
+            self.root_dir = self.buildDefaultRootDir()
+
+        self.default_model = self.getModelData('default')
+
+    def getSmartSliceData(self, metadata_path: str) -> Tuple[any, any, any, any]:
```
bradymadams, line 26: This will probably result in an error. `any` isn't a standard Python keyword and it's not imported from anywhere. What you're probably trying to use is `Any` from the `typing` module. However, use `Optional` instead if you know what the type is.

SmartSlicePlugin/tests/ModelLoader.py
```diff
@@ -0,0 +1,50 @@
...
+            metadata = json.load(f)
+
+        status = metadata.get("status", None)
+        job_dict = metadata.get("job", None)
+        results_dict = metadata.get("results", None)
+        row = metadata.get("selectedResult", None)  # The row is stored as the order of the results
+
+        return status, job_dict, results_dict, row
+
+    def getModelData(self, model_dir_path: str) -> Optional[T]:
```
bradymadams, line 37: When I wrote `Optional[T]` in a previous comment I wrote `T` to represent the type that the return is, if it's not None. `T` is commonly used as a generic placeholder. So if the return could be a string or None then the return type is `Optional[str]`.

It's important to use the proper types because good IDEs will pick it up and give you appropriate error checking and auto-completion.

bradymadams (changes requested):

SmartSlicePlugin/tests/ModelLoader.py
bradymadams: And make sure you set the return type back to a `Tuple`

SmartSlicePlugin/tests/ModelLoader.py
```diff
@@ -0,0 +1,50 @@
...
+        full_model_dir_path = os.path.join(self.root_dir, model_dir_path)
+        full_stl_path: str = os.path.join(full_model_dir_path, 'model.stl')
+        full_metadata_path = os.path.join(full_model_dir_path, 'plugin_metadata.json')
+
+        if not os.path.isfile(full_stl_path):
+            return None, None
+
+        if os.path.isfile(full_metadata_path):
+            smartslice_data: Tuple[any, any, any, any] = self.getSmartSliceData(full_metadata_path)
```
bradymadams, line 46: There's no need for the local annotation here since `getSmartSliceData` is annotated with the return type.

SmartSlicePlugin/tests/ModelLoader.py
AijunHall: Colman just linked me to a VSCode extension that does the type def checking, so I'll get that setup.

AijunHall:
I realized the selectedRow is also an int, so I reflected that in the new PluginDataType

colman-hartinger (approved):
Everything looks good to me

bradymadams (approved):
Looks good. I didn't notice this earlier, but note that one of the style exceptions in the Cura plug-in is we use double quote strings instead of single quote. I'll merge this anyway and we can update it in a later PR when someone comes across it.

PR #489, opened 2021-03-18
R21.0 fix plugin urls

AijunHall:
> We'll need to make sure this change doesn't break any other links if the concatenated routes don't contain the slash. In other words - let's make sure all URL creation is consistent with it's use of slashes.

I'm pretty sure this is the only place where a SmartSlice link is created with concatenation. 

The only two places where a SmartSlice link isn't explicitly written out in full `('https://www.tetonsim/...')` are in:

- `HelpTool.qml` with `Qt.openUrlExternally(proxy.accountUrl)`

- `SmartSliceLogin.qml` with `Qt.openUrlExternally(smartSliceMain.proxy.accountUrl + '/forgot')`

All the SmartSlice urls I've seen in the code (besides the 2 I've changed) never have a trailing slash


AijunHall:
Also the default PluginMetaData account url doesn't have a trailing slash

https://github.com/tetonsim/is-cura-ui/blob/39c81dee5f71cd14731d59ef303dedb33117aadd/SmartSlicePlugin/SmartSliceExtension.py#L312-L318

dougkenik (approved):

bradymadams (approved):

PR #496, opened 2021-04-01
fixed the mod mesh scaling bug

### LINKED ISSUE #471: Re-assigned mod mesh is scaled by parent's scale
See attached GIF. We move parent-less mod meshes to be a child of the printable mesh when the Smart Slice stage is selected. If the printable mesh was scaled then the mod mesh inherits this scale and it's very unlikely the user would want this behavior.

mod-mesh-scaled

dougkenik:
Fixes #471


colman-hartinger (approved):
looks good to me

bradymadams (approved):

AijunHall (approved):
Works good

Issue #494, opened 2021-03-30
Perform face selection in UM.Job


bradymadams:
There are many simple cases where the UM.Job could make the experience worse. So we may need to preemptively judge if the selection should be done in a Job or not.

PR #500, opened 2021-04-11
R21.0 surface job

dougkenik:
#494


SmartSlicePlugin/SmartSlicePreferences.py
```diff
@@ -9,6 +9,7 @@ class Preferences():
     Name = "first_name"
     SaveDebug = "debug_save_smartslice_package"
     DebugLocation = "debug_save_smartslice_package_location"
+    MaxNodes = "max_nodes"
```
bradymadams, line 12: Since this is referring to STLs vertices would be better nomenclature, IMO, to avoid confusion with FEA models.

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -170,31 +169,73 @@ def _getSelectedTriangles(
...
+            return
+
+        stage = self._controller.getActiveStage()
+
+        angle = bc_node.surface_angle * math.pi / 180.0
+
+        clamped = bc_node.surface_selection == SmartSliceScene.HighlightFace.SurfaceSelectionMode.Custom and \
+                  bc_node.surface_criteria == SmartSliceScene.HighlightFace.SurfaceSelectionCriteria.BySelection
+
+        if bc_node.surface_type == SmartSliceScene.HighlightFace.SurfaceType.Any and len(interactive_mesh.vertices) > self.extension.preferences.getPreference(Preferences.MaxNodes):
```
bradymadams, line 181: Break up long line.

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -170,31 +169,73 @@ def _getSelectedTriangles(
...
+
+            if self._calculating_surface_job and self._calculating_surface_job.isRunning():
+                self._calculating_surface_job.cancel()
```
bradymadams, line 184: We might need to look into this. As far as I can tell this won't kill the Job if it's already running. So that means if the `JobQueue` contains multiple workers it becomes a race between the existing SurfaceSelectionJob and the new SurfaceSelectionJob and whichever ones finishes last will determine the selected surface the user gets.

And consider if the user switched from surface type of `Any` to something else while the SurfaceSelectionJob is running. The non Any surface will probably select quickly and then the SurfaceSelectionJob will finish later and update the selected surface.

I think one fix would be to (before line 181) check if any SurfaceSelectionJobs are running and if so set a flag in them that they have been "cancelled". They will still need to run to completion (unless we made checks inside the surface selection code for a cancel request), but when the Job finished event signal gets hit it can check if the Job was cancelled or not.

bradymadams (changes requested):

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
```diff
@@ -450,3 +495,18 @@ def _changeRenderMode(self, faces=True):
...
+        self.interactive_mesh = interactive_mesh
+        self.face_id = face_id
+        self.node = bc_node
+        self.max_angle = max_angle
+        self.clamped = clamped
+        self.surface = None
+        self.canceled = False
+
+    def run(self):
+        time.sleep(1)
```
bradymadams, line 511: Let's double check if this is necessary or if we can reduce it. I think 1 second on the mesh analysis was fine because it's pretty trivial compared to the one time cost of analyzing the mesh, but for selecting surfaces over and over it might hurt the UX.

If we do need a sleep, let's add a quick comment explaining why. A random `sleep` call usually means we're improperly trying to fix a race condition which isn't necessarily the case here, so let's help the future dev out by explaining what it's for.

bradymadams (changes requested):

SmartSlicePlugin/select_tool/SmartSliceSelectTool.py
bradymadams: I don't know if I tried this in the mesh analysis Job, but we might also try calling `Job.yieldThread` which is just a wrapper around `time.sleep(0)`. That might be enough to yield to the UI thread to allow it to show the progress message box.

colman-hartinger (approved):
Looks good to me

bradymadams (approved):
Works nicely. Tried to break it, but couldn't.

Issue #491, opened 2021-03-23
Provide link to failure resolution if available.


dougkenik:
Right now, users do not have a path to resolution when an error is returned from our system. We should provide a link to a "Error Code" page where users can work on their own to fix it.

PR #503, opened 2021-04-14
R21.0 link consolidation

dougkenik:
Created a class which holds all of the links for SmartSlice, and registered a QML type so any UI component can access it. Modified all of our code so it uses the new class.


dougkenik:
Also added the work for #491, since it is closely tied to this

colman-hartinger (approved):
looks and runs good for me

SmartSlicePlugin/SmartSliceAbout.qml
```diff
@@ -105,7 +108,10 @@ UM.Dialog {
                         anchors.fill: parent
                         hoverEnabled: true
                         cursorShape: Qt.PointingHandCursor
-                        onClicked: Qt.openUrlExternally('https://help.tetonsim.com/eula-end-user-license-agreement')
+                        onClicked: {
+                            console.log("EULA: ", aboutDialog.urlHandler.eula)
```
bradymadams, line 112: Do we want this log?

SmartSlicePlugin/SmartSliceURL.py
```diff
@@ -0,0 +1,80 @@
...
+
+from .stage.SmartSliceStage import SmartSliceStage
+
+class URLHandler(QObject):
+    def __init__(self, parent=None):
+        super().__init__()
+
+        self._home = 'https://tetonsim.com'
+        self._help = 'https://help.tetonsim.com'
+        self._help_email = 'mailto:help@tetonsim.com'
```
bradymadams, line 12: Double quote strings. Same for all below strings in this file.

bradymadams (changes requested):

bradymadams (approved):

PR #504, opened 2021-04-16
made the face dialog appear moveable, and also fixed the forgot password link in the Login dialog

dougkenik:
The face selection dialog now appears to be moveable, and I fixed the forgot password link in the Login dialog


bradymadams (approved):

colman-hartinger (changes requested):
I think this may be from a  previous PR but the load magnitude helper is turned off in normal mode right now. Besides that the new handles look good and make it a lot clearer that the windows can be moved.

dougkenik:
> I think this may be from a previous PR but the load magnitude helper is turned off in normal mode right now. Besides that the new handles look good and make it a lot clearer that the windows can be moved.

This was definitely introduced by this PR... somehow. I think it is fixed now

colman-hartinger (approved):

Issue #509, opened 2021-04-16
Results buttons are off center in Cura 4.9


dougkenik:
See screenshot:

![image.png](https://images.zenhubusercontent.com/5ef3733ee554441fdbf0d022/2421809e-24e1-4f8a-be6e-9a38fc81eab2)

PR #510, opened 2021-04-16
R21.0 qml warnings

dougkenik:
This is a fix for #509 - I also cleaned up all of the QML code so we can get rid of the warnings in Cura


bradymadams (approved):

bradymadams:
@colman-hartinger can you take a look at this today?

colman-hartinger (approved):

Issue #459, opened 2021-02-17
Load arrow doesn't return to original position on "Cancel"


rickdalgarno:
When running a Validation, if you move the load arrow via the manipulation wheel, it will trigger a "Cancel/Continue" option for the user. If you click Cancel, the arrow does not return to the original position.

Steps:

1. Define a load where the load manipulation wheel is present.
2. Start a Validate.
3. During the Validate solution, adjust the load angle arrow. When prompted to Cancel or Continue, click Cancel. The arrow does not return to the original position.

Expected: the arrow returns to the original position after clicking Cancel.

See video below.

Tested with Cura 4.7 and r20.2 RC6.

https://user-images.githubusercontent.com/34459028/108270501-c7d42180-712c-11eb-8540-c41fd02cef07.mp4

[UMS5_shelf bracket_wheel_cancel.zip](https://github.com/tetonsim/is-cura-ui/files/5998721/UMS5_shelf.bracket_wheel_cancel.zip)

rickdalgarno:
IMO, this is not critical and can be pushed to a later release.

bradymadams:
I can repro this. It's definitely strange behavior. It happens about 1/2 the time, and if you move and cancel it several times it will eventually revert back to the original, correct position. We'll look into fixing this for the 20.2 release if something else critical comes up that we need to fix.

PR #515, opened 2021-04-23
R21.0 incorrect validation arrow direction bugfix

AijunHall:
Was a tough one for me to diagnose.

`numpy.arccos()` was taking arguments slightly outside its domain from the file in #459 and returning NaN incorrectly for certain angles.

`dot / denom` would sometimes come out as values like 1.0000000000000002 or -1.0000000000000002, so I am trying to deal with those outlying situations with this rounding


dougkenik (approved):
I am extremely impressed with this Aijun. Great job

bradymadams (approved):
These types of bugs can be really hard to track down. Really nice job @AijunHall.

Issue #522, opened 2021-04-29
Deselecting model during validation causes "Lose results" popup


bradymadams:
Not sure if this is unique to the model, custom surface selection or what, but there's two issues here. One, just deselecting the model causes the "lose results?" popup to come up. And clicking cancel and then reselecting the model causes it to come up again.

The second issue can also be seen in the gif if you watch the load arrow. After clicking cancel the arrow direction changes (both graphically and the vector components change).

[PI3MK2_Dog_Butt_Wall_Hanging_Final.zip](https://github.com/tetonsim/is-cura-ui/files/6401519/PI3MK2_Dog_Butt_Wall_Hanging_Final.zip)

![custom-load-arrow-cancel](https://user-images.githubusercontent.com/23110600/116615438-89ab6900-a8f8-11eb-8c2d-8f63815473cf.gif)

dougkenik:
I think this is related to #518

bradymadams:
> I think this is related to #518

@dougkenik did you mean #519?

dougkenik:
Yes - apologies

PR #530, opened 2021-05-02
fix for 522

dougkenik:
Fix for #522


SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -529,13 +529,36 @@ def setArrow(self, direction: Vector):
...
+
+            # Project the vector into the phi plane
+            phi_vector = direction - direction.dot(self.phiRotator.rotation_axis.normalized()) * self.phiRotator.rotation_axis.normalized()
+
+            angle = angleBetweenVectors(self.activeArrow.direction, phi_vector)
+
+            axes_angle = angleBetweenVectors(self.phiRotator.rotation_axis, phi_vector.cross(self.activeArrow.direction))
+            angle = -angle if abs(axes_angle) < 1.e-3 else angle
+
+            self.rotateArrow(angle, RotationAxis.Phi)
```
bradymadams, line 560: Lines 550-560 are repeated a few times. Wouldn't mind seeing those eventually consolidated.

bradymadams (approved):
It is behaving as expected now. The load arrows remain after clicking off the part, but that's because the fixes in #527 aren't in this PR.

rickdalgarno:
@dougkenik The load wheels still have some issues:

1. Define custom load and move both wheels.
2. During validation, adjust the angle slider, and click "Cancel" when prompted.

After clicking Cancel, the wheels re-orient in a seemingly random fashion. The load arrow, however, does appear to maintain the same direction.

I tested using Cura 4.7 and 4.9.

dougkenik:
> @dougkenik The load wheels still have some issues:
> 
> 1. Define custom load and move both wheels.
> 2. During validation, adjust the angle slider, and click "Cancel" when prompted.
> 
> After clicking Cancel, the wheels re-orient in a seemingly random fashion. The load arrow, however, does appear to maintain the same direction.
> 
> I tested using Cura 4.7 and 4.9.
> 
> 

@rickdalgarno - if the arrow returns to the original direction, then I would say we are good. We do not record the steps the user took to get to that orientation, so it is near impossible to put the rotators in the same position.

rickdalgarno (approved):
Tested using Cura 4.7 and 4.9 and build SmartSlicePlugin-r21.0-522.29. This fixes the issues in #522.

Issue #520, opened 2021-04-29
Deleting normal mesh after optimized and cancelling results in incorrectly placed mod mesh


bradymadams:
1. Rotate a part
2. Setup load case, validate, and optimize
3. Select a result with a mod mesh
4. Delete the normal mesh from the SmartSlice stage
5. Click cancel

The mod mesh is re-added, but the position is incorrect, and there are two mod meshes. [I doubt it's unique to the geometry/job I used, but here's the optimized 3MF just in case.](https://github.com/tetonsim/is-cura-ui/files/6401367/PI3MK2_floating-bracket-inverse.3mf.zip).

![mod-mesh-incorrect-undo](https://user-images.githubusercontent.com/23110600/116612296-9f1e9400-a8f4-11eb-8a15-a88c00bcbf56.gif)

PR #531, opened 2021-05-02
R21.0 520

dougkenik:
Fix for #520


SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -582,6 +591,10 @@ def restore(self):
         if not self.highlight_face.isVisible():
             self.highlight_face.disableTools()
 
+class SmartSliceLoadTool(TracketProperty):
+    def __init__(self):
+        self._theta_rotation_axis
+
```
bradymadams, line 597: SmartSlice isn't loading. I assume it's because the base class on line 594 is misspelled. Is this class even necessary for this PR?

bradymadams (changes requested):

SmartSlicePlugin/SmartSliceProperty.py
dougkenik: Nope

SmartSlicePlugin/SmartSliceProperty.py
```diff
@@ -411,15 +411,24 @@ def cache(self):
...
             parent = self._parents[n] if self._parents[n] else self._root
-            parent.addChild(n)
+
+            if not scene.findObject(id(parent)):
+                self._root.addChild(parent)
+
+            if n not in parent.getChildren():
+                parent.addChild(n)
+                n.rotate(n.getWorldOrientation(), 3)
+                n.scale(n.getWorldScale(), 3)
```
bradymadams, line 431: Let's use the `TransformSpace.World` variable on these two calls. It's more explicit and won't break if those variables ever change for some reason.

bradymadams (changes requested):
Bug is fixed. Just one code change request.

SmartSlicePlugin/SmartSliceProperty.py
dougkenik: I've tried to do that, but Python crashes and tells me `SceneNode` has no attribute `TransformSpace` - even though I have explicitly imported `SceneNode`

dougkenik: Nevermind - I see why now

bradymadams:
Out of curiosity - why did you have to rename `SceneNode` to `UMSceneNode`?

dougkenik:
> Out of curiosity - why did you have to rename `SceneNode` to `UMSceneNode`?

Because there is a class in that file called SceneNode. I think a better long term solution is to rename that class to something else, but this was the easier solution for the release

bradymadams (approved):

colman-hartinger (approved):

PR #536, opened 2021-05-06
Active machine none on Cura 4.9 shutdown

bradymadams:
Cura 4.9 sets the active machine to None while it's closing. This was causing a whole slew of CRITICAL errors to print out from our plugin. This resolves that by checking for a None activeMachine.


dougkenik (approved):

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -237,9 +250,7 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
                 "Enter a valid list of an angle (e.g. [45])"
             ))
 
-        print_config.auxiliary = self._getAuxDict(
-            Application.getInstance().getGlobalContainerStack()
-        )
+        print_config.auxiliary = self._getAuxDict(app.getGlobalContainerStack())
```
AijunHall, line 253: `Application` previously was `UM.Application` and now its been changed to the Cura application here. Is that ok? I'm curious about the difference between `getInstance()` from UM vs Cura

bradymadams: UM.Application is the generic class for an application for anything that wants to use the Uranium framework. CuraApplication is the Cura implementation of UM.Application. Honestly, there's no real good reason for our plug-in to ever use `UM.Application.getInstance()`, since our plug-in is made specifically for Cura - it will never be used in a different Uranium application. `UM.Application.getInstance()` will return the same application instance as `CuraApplication.getInstance()`. You would want to strictly stick to using the former if you were building something that is for the Uranium framework, and not Cura specifically.

AijunHall (approved):
Thanks for the explanation

PR #550, opened 2021-05-14
use the root node to write 3MFs so groups are written correctly

bradymadams:
[Build](https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/SmartSlicePlugin-r21.0-3mf-groups.43.curapackage)

This fixes an issue with writing groups to 3MFs. Groups occur when a mod mesh is assigned as a child to the printable mesh. We had two problems with how we were previously writing the nodes to the 3MF:

1. We were explicitly adding mod meshes to the list of nodes to write. Since the mod meshes were already children of the printable mesh they got included twice.
2. I don't know the exact details of why, but the transformations were not written as we'd expect when the printable node of a group was supplied directly to the 3MF writer. Instead, I've change it to write the 3MF as Cura does by providing the scene root. This appears to fix the transformation issues (@jlselden I added you to this ticket to double check this since you were helping me look at this yesterday).


bradymadams:
Also, I changed to using the Workspace writer which will put the Cura folder into the 3MF. It also writes the plugin metadata, but this is a little problematic to use - Our backend 3MF parser doesn't recognize it and there's a timing issue with the data. Rather than have all that data duplicated. I remove the SmartSlicePlugin folder from the 3MF, and I write the job.json to it's usual location.

The advantage here is for troubleshooting 3MFs for our solver we can use a simple script (I already have one I can share) that moves the Job data into the plugin metadata and the Cura folder will now be there so when we open it in Cura the full configuration will be loaded properly.

dougkenik (approved):

jlselden (approved):
Checks out. No duplicate modifier meshes and the transformations all look correct.

colman-hartinger (approved):
Looks good

Issue #551, opened 2021-05-14
Selecting faces on grouped scene nodes is faulty


bradymadams:
[two-cubes.3mf.zip](https://github.com/tetonsim/is-cura-ui/files/6479402/two-cubes.3mf.zip)

1. Open the attached 3MF
2. Go to the SmartSlice stage
3. Selecting faces for any BC doesn't work. The behavior appears to be different between Anchors and Loads, but neither works.

bradymadams:
We're going to release plug-in version 21.0 with this as a known issue.

There are two workarounds:
1. Ungroup the models before entering the SmartSlice stage. This workaround will hold until the 3MF is re-saved and/or re-opened.
2. Or hold CTRL while selecting faces in the SmartSlice stage.

PR #552, opened 2021-05-14
Z Offset Mod Meshes

bradymadams:
This is the start of trying to work out #551. I decided to try removing the functionality to group the mod mesh as a child of the printable node. To counteract Cura moving the mod mesh to the build plate if the _Drop parts to build plate_ setting is on I add a ZOffsetDecorator to mod meshes when a 3MF is loaded.

Cura removes the ZOffsetDecorator if the mod mesh is repositioned, but I think that's okay. The user can go turn off that setting if they need to move it off the build plate.

The one thing that's still problematic is the mod meshes returned from optimization. I think we're putting those as children when they come back, so that will have to change. But I wanted to open a draft PR first in case there's something I'm missing here that makes this solution infeasible. I know some of this had been explored in the past, so there might be a workflow that blows up that I've forgotten about.


SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -376,32 +372,19 @@ def _checkScene(self):
...
-            for intersecting_node in intersectingNodes(node):
-                if intersecting_node.getParent() != node:
-                    Tp = node.getWorldTransformation()
-                    Tc = intersecting_node.getWorldTransformation()
-                    intersecting_node.setParent(node)
-                    intersecting_node.setTransformation(Tp.getInverse().multiply(Tc))
-                    parent_changed = True
+        for n in nodes:
+            aabb = n.getBoundingBox()
+            if aabb.bottom > 0 and not n.hasDecoration("getZOffset"):
```
dougkenik, line 383: There were times when a mesh was read in, and it was under the build plate. Should we be checking for that, too?

dougkenik (changes requested):

SmartSlicePlugin/stage/SmartSliceStage.py
bradymadams: I was mostly thinking about the mod meshes we create, but I suppose one could validate with a mod mesh below the build plate, so I'll add that check.

bradymadams: Turns out Cura already adds the ZOffsetDecorator in PlatformPhysics.py if the mesh is below the build plate. I added the check anyway, but it won't trigger in our code because `node.hasDecoration("getZOffset")` will be true in those cases.

bradymadams:
@dougkenik I think this is ready for a look if you don't think there's any obvious problems here.

SmartSlicePlugin/stage/SmartSliceStage.py
dougkenik: So this is exactly the problem we ran into. Jeff was running a validation with a mod mesh, and then saving it. When it was read back in on project open, Cura put it under the build plate for some reason and added the decorator, and we had no way to tell where the mod mesh should actually be, so we had to move over to writing the mod mesh as a child of the normal mesh

bradymadams: I think we should test it cause I didn't see that behavior. And cura only adds the decorator if the mesh is already under the build plate, which keeps the PlatformPhysics from moving it. I just don't know why they use different logic for parts above the build plate. 

I've tested with 3mfs saved with mod meshes above and below the plate and they all seem to work for me (cura 4.9).

dougkenik: Ok - this is the issue where we saw the problems last time:

https://app.zenhub.com/workspaces/smartslice-5ef3719d317b88001a6dc21e/issues/tetonsim/is-cura-ui/297

bradymadams: Okay, I'll take a look at that specific case.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -376,32 +374,19 @@ def _checkScene(self):
...
-                if intersecting_node.getParent() != node:
-                    Tp = node.getWorldTransformation()
-                    Tc = intersecting_node.getWorldTransformation()
-                    intersecting_node.setParent(node)
-                    intersecting_node.setTransformation(Tp.getInverse().multiply(Tc))
-                    parent_changed = True
+        for n in nodes:
+            aabb = n.getBoundingBox()
+            if abs(aabb.bottom) > 1.e-5 and not n.hasDecoration("getZOffset"):
+                Logger.info("Adding ZOffsetDecorator (%f) to %s", aabb.bottom, n.getName())
```
colman-hartinger, line 386: It seems like for the most part we use string.format in this repo. I'm not sure if we want to stick with that

bradymadams:
@colman-hartinger you can hold off on a review here. I need to check a few cases and might need to make additional changes.

colman-hartinger (commented):
> @colman-hartinger you can hold off on a review here. I need to check a few cases and might need to make additional changes.

Sounds good. I had a few minor styling requests in the review I was doing so I'll just post those

SmartSlicePlugin/stage/SmartSliceStage.py
bradymadams: I didn't realize we were doing that. I'd suggest we go with this style. This is the standard use for log statements because the logger will skip formatting the string if the log level is lower than the level of the log statement. When using string format directly the string will always be formatted regardless if it's actually logged or not. So it's good practice to let the log library determine if the formatting needs to be done or not to keep performance as high as possible. (It's not going to make a difference in this case, but if it's a question of which style is preferred - I'd say this one is preferred).

bradymadams:
Not going to use this functionality. It's too problematic not to assign mod meshes as a child to the normal mesh. When loading the 3MF Cura will move around every node in the root, freely, into an "ideal" position (ThreeMFReader.py - `_read` function).

Issue #478, opened 2021-03-07
Don't let users enter the SmartSlice stage with non-manifold STL's


dougkenik:
Cura issues a warning to the user when they have non-manifold STL's. We should not let users enter the stage when this happens.

AijunHall:
- Added a check for non-manifold edges with improper connecting triangles to mesh analysis.

- The plugin will now issue a warning to users as they enter the SmartSliceStage with non-manifold geometry - as detected by the mesh analysis.

Going forward we can add different types of non-manifold checks, and choose how to best deal with those situations with more testing of ill-informed geometry.

PR #553, opened 2021-05-17
R21.0 non manifold detection

AijunHall:
This is expected to work off of my pywim PR: https://github.com/tetonsim/pywim/pull/87
That one would have to be approved and merged first to not cause problems.

For now I allow non-manifold geometry through with a warning instead of an exception, for ease of testing how often this situation will be brought up.


jlselden (changes requested):
Just some minor grammar stuff.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -655,6 +655,15 @@ def initialize(self, parent: SceneNode, max_vertices, step=None, callback=None):
                     self.setOrigin()
                 if callback:
                     callback()
+
+                if self._interactive_mesh.non_manifold_warning:
+                    self._mesh_non_manifold_warning_message = Message(
+                        title=i18n_catalog.i18n("SmartSlice has detected non manifold geometry"),
```
jlselden, line 661: Add a period.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -655,6 +655,15 @@ def initialize(self, parent: SceneNode, max_vertices, step=None, callback=None):
...
+                        text=i18n_catalog.i18n("This may affect the accuracy of anchors and loads"),
```
jlselden, line 662: Add a period.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -182,6 +184,16 @@ def onStageSelected(self):
                     printable_node,
                     self.extension.preferences.getPreference(self.extension.preferences.MaxVertices)
                 )
+
+            except pywim.geom.tri.NonManifold as exc:
+                Logger.logException("e", "Detected non manifold")
+                self._scene_not_ready(
+                    i18n_catalog.i18n("SmartSlice can not be used with non manifolds. Please remove the model")
```
jlselden, line 191: Add a period.

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -182,6 +184,16 @@ def onStageSelected(self):
...
+                Logger.logException("e", "Detected non manifold")
+                self._scene_not_ready(
+                    i18n_catalog.i18n("SmartSlice can not be used with non manifolds. Please remove the model")
+                )
+                if smart_slice_node:
+                    printable_node.removeChild(smart_slice_node)
+                return
+
             except Exception as exc:
                 Logger.logException("e", "Unable to analyze geometry")
```
jlselden, line 198: Fix Doug's sloppy grammar by adding a period.

AijunHall: Should all logExceptions and Message titles have periods?

SmartSlicePlugin/stage/SmartSliceStage.py
```diff
@@ -182,6 +184,16 @@ def onStageSelected(self):
                     printable_node,
                     self.extension.preferences.getPreference(self.extension.preferences.MaxVertices)
                 )
+
+            except pywim.geom.tri.NonManifold as exc:
```
bradymadams, line 188: You'll need to be careful here. Take a look at `SmartSliceNode.initialize`. If the geometry exceeds a certain size the mesh analysis gets offloaded into a `Job`. If an exception gets thrown in a `Job` it won't get caught here.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -655,6 +655,15 @@ def initialize(self, parent: SceneNode, max_vertices, step=None, callback=None):
                     self.setOrigin()
                 if callback:
                     callback()
+
+                if self._interactive_mesh.non_manifold_warning:
+                    self._mesh_non_manifold_warning_message = Message(
```
bradymadams, line 660: It doesn't look like `self._mesh_non_manifold_warning_message` is used outside of this function, therefore it doesn't need to be a class attribute.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -655,6 +655,15 @@ def initialize(self, parent: SceneNode, max_vertices, step=None, callback=None):
...
+                        title=i18n_catalog.i18n("SmartSlice has detected non manifold geometry"),
+                        text=i18n_catalog.i18n("This may affect the accuracy of anchors and loads"),
+                        lifetime=0,
+                    )
+                    self._mesh_non_manifold_warning_message.show()
+
             else:
                 job = AnalyzeMeshJob(mesh_data, step, callback)
```
bradymadams, line 668: I alluded to this in another comment, but we'll also need the same checks when the mesh analysis is done in the `AnalyzeMeshJob`.

bradymadams (changes requested):

SmartSlicePlugin/stage/SmartSliceStage.py
jlselden: In my opinion, if a customer is going to see it, it should be grammatically correct. I think it looks more professional that way. If it's only ever going to be seen by us, it's less important.

SmartSlicePlugin/stage/SmartSliceStage.py
bradymadams: Ping me if you need some help understanding the `Job` classes and how they work. They essentially offload a function into another thread so the UI doesn't lock up.

AijunHall: I see where I need to include another mesh warning check in the `Job` now, I just missed it since I didn't immediately see the interactive mesh in that else block.

SmartSlicePlugin/stage/SmartSliceScene.py
```diff
@@ -656,13 +656,13 @@ def initialize(self, parent: SceneNode, max_vertices, step=None, callback=None):
                 if callback:
                     callback()
 
-                if self._interactive_mesh.non_manifold_warning:
-                    self._mesh_non_manifold_warning_message = Message(
-                        title=i18n_catalog.i18n("SmartSlice has detected non manifold geometry"),
-                        text=i18n_catalog.i18n("This may affect the accuracy of anchors and loads"),
+                if any(issubclass(type(warning), pywim.geom.tri.NonManifold) for warning in self._interactive_mesh.warnings()):
```
bradymadams, line 659: When you have an object `isinstance` is preferred over `issubclass`:

```python
isinstance(warning, pywim.geom.tri.NonManifold)
```

bradymadams (changes requested):
Looks pretty good. Just one minor code change request. Can you share a screenshot showing how the message looks?

AijunHall:
Heres how the warning message looks @bradymadams .

image

But don't I still have to include an exception for potential queued Jobs? You said above that the except catch on line 188 in SmartSliceStage won't catch Job.Errors

I was still working on putting the same exception catch here 

https://github.com/tetonsim/is-cura-ui/blob/0d3c3b4823a18f91027a095f600c652adbdf74d3/SmartSlicePlugin/stage/SmartSliceScene.py#L681-L689

bradymadams:
> But don't I still have to include an exception for potential queued Jobs? You said above that the except catch on line 188 in SmartSliceStage won't catch Job.Errors
> 
> I was still working on putting the same exception catch here
> 
> https://github.com/tetonsim/is-cura-ui/blob/0d3c3b4823a18f91027a095f600c652adbdf74d3/SmartSlicePlugin/stage/SmartSliceScene.py#L681-L689

No, you're correct. I forgot about that - we need to catch the exception in the Job as well.

bradymadams:
I was working through a handful of PR reviews and didn't realize you hadn't re-requested my review on this PR yet - that's my bad.

bradymadams:
Let's change the warning message a little bit. It hard potentially further reaching issues than just the boundary conditions.

> SmartSlice has detected non manifold geometry.
> This may affect the accuracy of anchors and loads.

to

> Warning: the geometry is not manifold.
> Non-manifold geometry may cause unexpected behavior or errors when used with SmartSlice.

AijunHall:
I wana note that the `Job.Error()` won't force users out of the SmartSliceStage like it does for models of lesser sizes that don't get queued up in `initialize()` I don't know if that's something that needs to be fixed for #478 . That's if pywim has opted to raise Errors instead of storing them in the mesh warning.

The other already existing catchall exception for `Job.Errors()` doesn't kick users out of the SmartSliceStage either.

AijunHall:
> I see the issue with the exception in the Job not exiting the stage. I think rather than fix that, it probably makes sense to just remove the check for the exceptions so there's no confusion in the future. We can add the checks back in, and add the logic to exit the stage, if we ever decide we need to reject non-manifold geometries. Thoughts?

Sounds good to me, if more non-manifold checks get added in, the way we deal with these mesh warnings is probably going to change anyway.

AijunHall:
Only the warning exists now

jlselden (approved):
Looks good to me.

dougkenik (approved):

bradymadams (approved):
Looks good @AijunHall

PR #560, opened 2021-06-17
R21.1 mod mesh validation

AijunHall:
This PR works off my [other PR in pywim.](https://github.com/tetonsim/pywim/pull/89)

Here I changed the validation in `checkJob` to work per node/mesh, instead of only for the global print config. This way the Top/Bottom Line Directions, Infill Line Directions, and Infill pattern are all validated properly for mod meshes. For mod meshes I use the existing auxiliary dictionary to get properties vs changing the propertyHandler

Without the pywim changes `_getAuxDict` doesn't retrieve certain properties. For example previously:
```
for prop in pywim.smartslice.val.NECESSARY_PRINT_PARAMETERS:
            val = stack.getProperty(prop, "value")
```
infill.pattern, infill.orientation, infill.density all aren't recognized by Cura and return None which causes issues for validation.

An alternative to the pywim changes is putting checks in `_getAuxDict` that change the pywim infill property keys to their Cura equivalents (infill_pattern, infill_sparse_density, etc).


bradymadams:
Doug and I may not be able to review this for a couple of days, so adding Colman as a reviewer.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,118 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
+        extruderManager = app.getExtruderManager()
+        emActive = extruderManager._active_extruder_index
+
+        # Extruder config
+        extruders = ()
+        machine_extruder = getNodeActiveExtruder(normal_mesh)
+        for extruder_stack in [machine_extruder]:
+            extruder = pywim.chop.machine.Extruder(diameter=extruder_stack.getProperty("machine_nozzle_size", "value"))
+            extruder.print_config.auxiliary = self._getAuxDict(extruder_stack)
+            extruders += (extruder,)
```
colman-hartinger, line 120: In python I like to use list comprehensions to create lists of objects rather than create an empty object and add to it. I think a cleaner way could be something like this

```
def setExtruder(stack):
    extruder = pywim.chop.machine.Extruder(diameter=stack.getProperty("machine_nozzle_size", "value"))
    extruder.print_config.auxiliary = self._getAuxDict(stack)
    return extruder

machine_extruder = getNodeActiveExtruder(normal_mesh)
extruders = [setExtruder(extruder_Stack) for extruder_Stack in machine_extruder]
```

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,118 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
+            # Skin angles
+            parsed_skin_angles = []
+
+            if isinstance(skin_angles, str):
+                parsed_skin_angles = SettingFunction(skin_angles)(self._propertyHandler)
+
+            if isinstance(parsed_skin_angles, list):
+                if len(parsed_skin_angles) == 0:
+                    print_config.skin_orientations.extend((45, 135))
+
```
colman-hartinger, line 181: remove the empty line between if-else. Also instead of checking the length, you can just use `if not parsed_skin_angles`.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,118 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
+                else:
+                    print_config.skin_orientations.extend(tuple(parsed_skin_angles))
```
colman-hartinger, line 183: add empty line after if-else statement

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,118 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
+                errors.append(pywim.smartslice.val.InvalidSetup(
+                    "Invalid <i>Top/Bottom Line Directions</i> for mesh <i>{}</i>: {}".format(mesh.name, skin_angles),
+                    "Enter a valid list of angles (e.g. [45, 135])"
+                ))
+
+            # Infill angles
+            parsed_infill_angles = []
+
+            if isinstance(infill_angles, str):
+                parsed_infill_angles = SettingFunction(infill_angles)(self._propertyHandler)
```
colman-hartinger, line 194: Does the `SettingFunctoin` return a list? If it doesn't I don't think mixing data types is a good idea and could result in bugs down the line

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,118 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
+
+            if isinstance(parsed_infill_angles, list):
+                if len(parsed_infill_angles) == 0:
+                    print_config.infill.orientation = self.INFILL_DIRECTION
+
+                else:
+                    if len(parsed_infill_angles) > 1:
+                        Logger.log("w", "More than one infill angle is set! Only the first will be taken!")
+                        Logger.log("d", "Ignoring the angles: {}".format(parsed_infill_angles[1:]))
+                    print_config.infill.orientation = parsed_infill_angles[0]
```
colman-hartinger, line 204: Same  comment as above about no empty space between if-else statements and an empty line after them

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,118 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
+
+            if isinstance(infill_angles, str):
+                parsed_infill_angles = SettingFunction(infill_angles)(self._propertyHandler)
+
+            if isinstance(parsed_infill_angles, list):
+                if len(parsed_infill_angles) == 0:
+                    print_config.infill.orientation = self.INFILL_DIRECTION
+
+                else:
+                    if len(parsed_infill_angles) > 1:
```
colman-hartinger, line 201: I don't think this if statement is necessary since it's only used for logging info.  Maybe a better solution would be to just have a log statement in the parent else block saying "Setting infill angle <angle>. Ignoring other angles <other angles>"

colman-hartinger (changes requested):
Everything is looking pretty good, just some comments about styling and structure. This is just an initial code review so I'll test the functionality in Cura a bit later today

SmartSlicePlugin/SmartSliceJobHandler.py
AijunHall: Yeah `SettingFunction` is only either a list or None as far as I've seen.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,114 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
 
         # Get all nodes to cycle through
         nodes = [normal_mesh] + getModifierMeshes()
 
+        # Extruder Manager
+        extruderManager = app.getExtruderManager()
+        emActive = extruderManager._active_extruder_index
+
+        # Extruder config
+        extruders = ()
```
colman-hartinger, line 115: No need to initialize extruders here.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -101,31 +101,114 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
+            # > https://github.com/Ultimaker/CuraEngine/blob/master/src/FffGcodeWriter.cpp#L402
+            # > https://github.com/Ultimaker/CuraEngine/blob/master/src/FffGcodeWriter.cpp#L366
+
+            if mesh.type is pywim.chop.mesh.MeshType.normal:
+                print_config = global_print_config
+
+                skin_angles = self._propertyHandler.getExtruderProperty("skin_angles")
+                infill_angles = self._propertyHandler.getExtruderProperty("infill_angles")
+                infill_pattern = self._propertyHandler.getExtruderProperty("infill_pattern")
+
```
colman-hartinger, line 161: remove empty line between if-else statement

colman-hartinger (changes requested):

colman-hartinger (approved):
Code looks good and seems to fix associated bug

bradymadams (approved):
Code looks good and the issue is resolved. I did notice another similar issue, but it appears it was an existing issue, so I'll open a separate ticket for it. Nice job @AijunHall

Issue #541, opened 2021-05-07
Validate button remains sick after addressing item to be resolved


rickdalgarno:
Steps:
1. Replace the material database file with the file here (the issue is with a new tested material so the existing matDB file doesn't have the GUID): https://tetoncompositescom.sharepoint.com/:f:/s/TetonSimulation/El8VVKr-PNBHjk1UVgKtgBkB1l6GyIXbdd-yYHVxtYUq9A?e=EgK8Ye
2. Open Cura and setup a basic validation using generic ABS. 
3. Switch the material to Jabil PA0600 (will need to install from the Cura Marketplace). Notice the Validate button goes sick.
4. Resolve the issues in the Validate warning message.
5. The button remains sick.

If you change another print parameter (infill density, for example), then the button goes unsick.

https://user-images.githubusercontent.com/34459028/117508559-09af7f80-af46-11eb-8f9f-9494defbb9d7.mp4

dougkenik:
@rickdalgarno - Was this in 4.9?

rickdalgarno:
@dougkenik Roger that. I don't think it is a big deal as these print parameters are set to lines for nearly every other material in our database so this issue is more or less specific to this material. However, if we ever do a case study with Jabil on this material, then it would be good to fix this issue.

PR #565, opened 2021-06-29
R21.1 Jabil PA0600 validation fix

AijunHall:
Should fix [541](https://github.com/tetonsim/is-cura-ui/issues/541).

Since Jabil PA0600 has extruder property changes, the cache wasn't picking it up correctly between material changes.


dougkenik (approved):
@AijunHall - does this noticeably slow down the product at all when the material changes when compared to the previous version?

AijunHall:
> @AijunHall - does this noticeably slow down the product at all when the material changes when compared to the previous version?

Not noticeably. I just went now and tested the load times again, the difference to the previous version is negligible.

bradymadams (approved):

bradymadams:
@rickdalgarno since you opened #541 can you check this? Here's the build:

https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/SmartSlicePlugin-r21.1-JabilPA0600-validation-fix.1.curapackage

PR #567, opened 2021-07-07
R21.1 settingfunctions for print config

AijunHall:
`getRawProperty()` returns the SettingFunction if the user doesn't specify the value, and from that we can extract the string expression.

Here's some examples of what these expressions look like in the print config, stored as the cached value:
```
bottom_layers = "999999 if infill_sparse_density == 100 else math.ceil(round(bottom_thickness / resolveOrValue('layer_height'), 4))"

infill_sparse_thickness = "resolveOrValue('layer_height')"

infill_line_width = "round(line_width * 0.42 / 0.35, 2)"
```

I'm not sure if changes should be included in pywim to account for these print config expressions, or if thats entirely the responsibility of the backend to evaluate. I say this because currently this branch doesn't allow the user to submit any jobs since the validation is held up by these unaccounted for expressions. Let me know if there's more steps that should be taken for this on the plugin end.


AijunHall:
One more thing I want to note about some extruder properties, when I peep into the debugger to see their SettingFunction expressions, they aren't what I expect them to be.

I've been using `Cura\resources\definitions\fdmprinter.def.json` as a reference for what the expressions should be.

For example in said `fdmprinter.def.json`, 
```
settings
{
  resolution
  {
    children
    {
      line_width
      {
         ... ,
         ... ,
         "value": "machine_nozzle_size",
         ... ,
      }
    }
  }
}
```
line width has a value of `"machine_nozzle_size"`

But when I peep into the debugger at the extruder line_width value within our SmartSlicePropertyHandler, the expression comes out as `"machine_nozzle_size * 0.875"`. I can provide a few more examples, but for the most part the expressions match up. 

Is this the result of loading different models with preset profiles and settings, or is this a concern with using the `getRawProperty()` method?

bradymadams:
`fdmprinter.def.json` is the base/default settings. You might be seeing a setting that is overridden by the selected printer's profile. You could probably find it's related JSON file in the same folder as `fdmprinter.def.json` and inspect it to confirm or deny that theory.

AijunHall:
> `fdmprinter.def.json` is the base/default settings. You might be seeing a setting that is overridden by the selected printer's profile. You could probably find it's related JSON file in the same folder as `fdmprinter.def.json` and inspect it to confirm or deny that theory.

I found the relevant printer json profile with the corresponding values, so those check out right. Thanks for clearing that up for me.

bradymadams (changes requested):
@AijunHall I think this looks pretty good. Nice job finding where to get the function string. A couple thoughts:

1. Looks like the code you added is identical in all three cases. Can we consolidate it into a function?
2. We should avoid accessing "private" class attributes as best we can (`_code` in this case). I'm assuming there was no other way to get this value?

We won't be able to merge this anytime soon as there's much more work on the back-end that will need to get done before this can be supported. So this may sit stagnant for awhile. Can you generate a 3MF for a small simple model and upload it to this PR?

AijunHall:
> 1. Looks like the code you added is identical in all three cases. Can we consolidate it into a function?

Putting this consolidated function in the utils seemed the most appropriate to me.

> 2. We should avoid accessing "private" class attributes as best we can (`_code` in this case). I'm assuming there was no other way to get this value?

Not that I'm aware of. SettingFunctions do have a `getUsedSettingKeys()` method, but that only returns the property keys involved in the expressions, not the full expressions themselves. 

I believe `_code` is the only way to retrieve the string.

AijunHall:
Heres that small simple model 3MF: 

[567_cube.zip](https://github.com/tetonsim/is-cura-ui/files/6804677/567_cube.zip)

bradymadams (approved):
Looks good. But like I mentioned we'll leave this unmerged until further work is done on downstream code to handle these expressions.

bradymadams:
Marking as a draft temporarily to turn off notifications until we're ready to merge this.

Issue #561, opened 2021-06-24
Respect print settings hierarchy on SmartSlice job creation and setting validation


bradymadams:
When we create the job in SmartSliceJobHandler.py and validate the settings we don't accurately respect the hierarchy of print settings (mesh -> extruder -> global). For example, set the mesh infill pattern to an unsupported pattern and SmartSlice won't catch it. It may be beneficial to take a step back and implement a more robust system for storing and retrieving settings in our own system, rather than just quickly patching this one scenario.

![image](https://user-images.githubusercontent.com/23110600/123273571-ffc6e900-d4bf-11eb-8f66-9daaad967e0b.png)

Related to #554

AijunHall:
> For example, set the mesh infill pattern to an unsupported pattern and SmartSlice won't catch it.

The changes made in [this previous PR](https://github.com/tetonsim/is-cura-ui/pull/560) made the plugin job validation happen per node, so at least on the Cura plugin side, this should now be caught.

I forgot to mention that there still might be some back-end issues with individual mesh print configs, as in, how these individual mesh print config settings are being saved to the solver might still be an issue. 

Let me know what the situation is after your testing

PR #574, opened 2021-08-04
R21.1 checkJob code consolidation

AijunHall:
Opting to upload this as a draft to get eyes on this sooner as I make progress. My goal is to retain the "top level resources" like the job, errors, and nodes while splitting other functionality into separate methods to try and be more readable and efficient. This is about 80% done I think. 

I still need to make changes to my proposed `_validateConfigs` and `_buildConfigs`. The `normal_mesh` needs to have its own individual mesh config and global config validated at the same time, as mentioned in [561](https://github.com/tetonsim/is-cura-ui/issues/561). I thought I'd managed that, but I still have some work to do. I also need to include typing declarations for these new methods.

Additionally `buildJobFor3mf` will have to be updated to reflect these new print hierarchy changes (including the individual mesh print configs alongside the global/extruder configs). I could either create a new ticket for that, or include the changes here.


SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -77,44 +77,75 @@ def __init__(self, handler: SmartSlicePropertyHandler):
     # Builds and checks a SmartSlice job for errors based on current setup defined by the property handler
     # Will return the job, and a dictionary of error keys and associated error resolutions
     def checkJob(self, writing_workspace=False, machine_name="printer", show_material_warnings=False) -> Tuple[pywim.smartslice.job.Job, Dict[str, str]]:
+        printable_nodes = getPrintableNodes()
 
-        if len(getPrintableNodes()) == 0:
+        if not printable_nodes:
```
dougkenik, line 82: Might want to keep the check for the length being zero here as well, just in case

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -77,44 +77,75 @@ def __init__(self, handler: SmartSlicePropertyHandler):
...
+                global_print_config.infill.orientation = infill_angles
+                global_print_config.infill.pattern = infill_pattern
 
-        # Extruder config
-        machine_extruder = [getNodeActiveExtruder(normal_mesh)]
-        extruders = [self._setExtruder(extruder_stack) for extruder_stack in machine_extruder]
+            job.chop.meshes.add(mesh)
+
+        # Use Cases
+        smart_sliceScene_node = findChildSceneNode(normal_node, Root)
```
dougkenik, line 125: Naming here is off - smartslice_scene_node might be better

dougkenik (commented):
In general this seems to be consolidating a tiny bit. One thing I would recommend is to return a set of new errors from your functions and add them to cura_errors instead of passing cura_errors around to them

AijunHall:
I was running into an issue where if a node had a valid mesh infill pattern, but an invalid extruder infill pattern, the error wouldn't be caught. I added a Infill pattern error creation in the global print config validation for now to fix that, rather than letting it go to pywim, as is done for the meshes. I'll look more into that to see if its necessary.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -205,8 +236,30 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
             # Infill pattern
             if infill_pattern in self.INFILL_CURA_SMARTSLICE.keys():
                 print_config.infill.pattern = self.INFILL_CURA_SMARTSLICE[infill_pattern]
+            elif i == 0:
+                errors.append(pywim.smartslice.val.InvalidSetup(
+                    "Unsupported <i>Infill Pattern</i> for mesh <i>{}</i>.".format(mesh.name),
+                    "<i>Infill Pattern</i> must be set to one of <i>[grid, triangle, cubic]</i>."
+                ))
```
bradymadams, line 246: Why is special handling of the normal mesh infill pattern necessary? If possible, I'd prefer we use a consolidated check in pywim.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -23,6 +23,7 @@
 
 from cura.Settings.ExtruderManager import ExtruderManager
 from cura.CuraApplication import CuraApplication
+from cura.Scene.CuraSceneNode import CuraSceneNode
```
AijunHall, line 26: When I was creating the cura node typing declarations for my other functions, I was running into issues with cura not being defined. This import is solely for the typing.

`all_nodes: List[CuraSceneNode]`

Is it appropriate to have an import purely for the typing, or should I be doing the typing differently?

SmartSlicePlugin/SmartSliceJobHandler.py
AijunHall: Looking into this right now, I agree that we should be using the consolidated check in pywim.

Might be able to get rid of this by adding the global print config to the job's meshes

SmartSlicePlugin/SmartSliceJobHandler.py
bradymadams: I'm fine with it. [Some code bases will add a conditional where the import is only done on type checking.](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING)

SmartSlicePlugin/SmartSliceJobHandler.py
AijunHall: @bradymadams Just tried this and it seems to work.

What do you think about adding an additional mesh to the job meshes that holds the global print config? That way the special handling here isn't necessary. 

One issue with this though is the pywim error message might be confusing to users, depending on what we name the mesh that holds the global print config
image

Another option I see is creating a method in pywim dedicated to validating the global print config separate from the existing `job.validate()` which works through the meshes. I could be missing an easier solution here though. What are your thoughts?

dougkenik: I would prefer to move all error checking into job.validate. I believe that function is setup to use the global print config whenever a mesh config is not set. If not, then we might need to work out what to do in that function to catch these errors. I also am uneasy adding the normal mesh again just to check the global print config... there should be a better solution for this.

dougkenik (changes requested):
This is looking much cleaner, and I like how we have broken out the major parts of the code base.

AijunHall:
I'll see what I can do to further consolidate this validation towards pywim and `job.validate()`

dougkenik:
@AijunHall - what do you mean by  "valid" and "invalid" in this context?

AijunHall:
> @AijunHall - what do you mean by "valid" and "invalid" in this context?

@dougkenik By valid, I mean a user input from the Top/Bottom Line Directions, Infill Directions, and Infill Pattern entry fields that is acceptable to enter into the print configs, and usable on the backend for the solver. The UM SettingFunction does some of this work in interpreting what is a "valid" or "invalid" user entry, and I try to translate that into our config.

By Invalid, I mean a user input that isn't acceptable/usable. For example we previously had crashes when users would enter a string list like ` 1, 2, 3]` into the Top/Bottom Line Directions or Infill Directions because of how the UM SettingFunction would translate that to our print config.

My pywim PR is trying to offload the functionality that makes that distinction of what is "valid" and "invalid"

dougkenik:
> > @AijunHall - what do you mean by "valid" and "invalid" in this context?
> 
> @dougkenik By valid, I mean a user input from the Top/Bottom Line Directions, Infill Directions, and Infill Pattern entry fields that is acceptable to enter into the print configs, and usable on the backend for the solver. The UM SettingFunction does some of this work in interpreting what is a "valid" or "invalid" user entry, and I try to translate that into our config.
> 
> By Invalid, I mean a user input that isn't acceptable/usable. For example we previously had crashes when users would enter a string list like ` 1, 2, 3]` into the Top/Bottom Line Directions or Infill Directions because of how the UM SettingFunction would translate that to our print config.
> 
> To be as simple as possible I really mean a valid list value, and an invalid list value
> 
> My pywim PR is trying to offload the functionality that makes that distinction of what is "valid" and "invalid"

Ok - so if it is invalid, shouldn't we be catching that prior to submitting the job? It should be caught as soon as it's input and invalid by the plugin job checking, shouldn't it?

AijunHall:
> Ok - so if it is invalid, shouldn't we be catching that prior to submitting the job? It should be caught as soon as it's input and invalid by the plugin job checking, shouldn't it?

@dougkenik Thats what we were doing previously here within `checkJob`, adding the errors as we caught them plugin-side instead of letting it go to `job.validate()`. We had special handling of these fields that I've tried to consolidate into pywim.

My question now is what part of the validation would we like to consolidate into pywim? Should we have plugin-side error creation for these fields as well as pywim-side validation? Or should I revert back to previous commits in this PR where I included special error handling before validating the job? Thats what my`_validateConfigs` function was doing before.

I've moved all print config error checking consolidation into pywim based off you and brady's comments above mentioning you'd both like to get rid of the plugin-side handling of errors. But if its more desirable to add it back, that works as well.

dougkenik (approved):

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -469,11 +488,16 @@ def _modifyInfillAnglesInSettingDict(self, settings):
         for key, value in settings.items():
             if key == "infill_angles":
                 if isinstance(value, str):
-                    value = SettingFunction(value)(self._propertyHandler)
-                if len(value) == 0:
-                    settings[key] = [self.INFILL_DIRECTION]
+                    parsed_value = SettingFunction(value)(self._propertyHandler)
+
+                if isinstance(parsed_value, list):
```
bradymadams, line 493: If line 490 evaluates to false then `parsed_value` will be undefined here.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -240,8 +274,15 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
                         "Change active adhesion extruders to Extruder 1"
                     ))
 
-        # Check the material
-        machine_extruder = getNodeActiveExtruder(normal_mesh)
+        return errors
+
+    def _validateMaterial(self, normal_node: CuraSceneNode, show_material_warnings: bool) -> Tuple[Optional[pywim.fea.model.Material], List[pywim.smartslice.val.InvalidSetup]]:
```
bradymadams, line 279: Break up long line. This line exceeds the width of the code pane in Github when reviewing the PR. Take a look at the code changes under this PR and you should see what I mean.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -145,68 +169,76 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
 
             # Infill pattern
             if infill_pattern in self.INFILL_CURA_SMARTSLICE.keys():
                 print_config.infill.pattern = self.INFILL_CURA_SMARTSLICE[infill_pattern]
-            else:
-                print_config.infill.pattern = infill_pattern # The job validation will handle the error
+            elif i != 0:
+                print_config.infill.pattern = infill_pattern
+
+    def _validateExtruders(self, all_nodes: List[CuraSceneNode], app: CuraApplication, show_material_warnings: bool) -> List[pywim.smartslice.val.InvalidSetup]:
```
bradymadams, line 226: Break up long line.

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -45,6 +46,8 @@
 
 """
 
+BuiltConfigs = Tuple[List[pywim.chop.mesh.Mesh], List[pywim.chop.machine.Extruder], pywim.am.Config]
```
bradymadams, line 49: Really nice use of a type alias :+1:

bradymadams (changes requested):
Just a few minor requests.

SmartSlicePlugin/SmartSliceJobHandler.py
AijunHall: Thanks for catching this

SmartSlicePlugin/SmartSliceJobHandler.py
```diff
@@ -145,68 +169,79 @@ def checkJob(self, writing_workspace=False, machine_name="printer", show_materia
...
                 print_config.infill.pattern = self.INFILL_CURA_SMARTSLICE[infill_pattern]
-            else:
-                print_config.infill.pattern = infill_pattern # The job validation will handle the error
+            elif i != 0:
+                print_config.infill.pattern = infill_pattern
+
+    def _validateExtruders(self,
+                           all_nodes: List[CuraSceneNode],
+                           app: CuraApplication,
+                           show_material_warnings: bool) -> List[pywim.smartslice.val.InvalidSetup]:
```
bradymadams, line 229: Our preferred multi-line style for argument lists and function calls is to only add one indent level for the arguments, and then place the closing brace/bracket at the same level of indent as the first line. For example:

```python
def _validateExtruders(
    self,
    all_nodes: List[CuraSceneNode],
    app: CuraApplication,
    show_material_warnings: bool
) -> List[pywim.smartslice.val.InvalidSetup]:
```

You might see the style you used elsewhere in this code base, but that would be code from before we adopted a more "official" style guide.

bradymadams (changes requested):
Just need a slight format adjustment to the multi-line function definitions.

bradymadams (approved):

Issue #564, opened 2021-06-28
3rd party login


bradymadams:
Implement 3rd party login. Available identity providers will need to be configurable and not hard-coded so it can be changed depending on which slicer the plugin is installed in.

Will also need to detect trial statuses and display a message on first login.

More UX details to come.

PR #575, opened 2021-08-10
OAuth login

bradymadams:
#564

This introduces OAuth login functionality for the plugin. This has two benefits:

1. Passwords are not necessary in the plugin for logging in which is a security concern on an open platform like Cura.
2. We can delegate authentication of users to 3rd parties much more easily (e.g. sign in with an Ultimaker account).

When the sign in button is clicked:

1. An http callback listener is started ([see related pywim PR](https://github.com/tetonsim/pywim/pull/93))
2. Browser is opened to a login form and the user is asked to login
3. If login is successful (or user is already logged in) the user is asked to login to the plugin. This will trigger an authorization code to be sent back to the callback listener started in step 1.
4. The callback receives the auth code and makes a final call to thor to check if it's valid and receive an authorization token in return.
5. User is now signed in.

A couple other things to note in this PR:

1. I left the basic authentication functionality in place, but it is configurable in plugin.json. Security audits will probably still pick up on this, so if we prefer the new OAuth login functionality we may want to remove the basic auth altogether.
2. There are scenarios where a localhost listener cannot be started. The pywim oauth handler will recognize this and indicate that it couldn't be started. We will need to handle this special case and create a dialog to have the user copy/paste a code after logging in. I did not include this functionality in this PR in the interest of time.
3. plugin.json now has several configuration parameters for authentication that get passed on to the `SmartSliceAuth.AuthConfiguration` class, including: turning on/off basic auth, turning on/off oauth, configuring the identity providers to display in the browser login form.
4. I did some cleanup and broke some classes out into their own files. Mainly I pulled classes out of SmartSliceCloudConnector.py. I felt that file was getting way too large and difficult to navigate.


bradymadams:
There's some QML and logic in here for authenticating against a 3rd party directly from the plugin rather than delegating it to the SmartSlice login page. I don't think we need this and I plan on removing it, but I want to talk to @dougkenik about it first to make sure I'm not missing any workflows we might need to support in the future.

bradymadams:
[A build targeting the test server](https://s3.console.aws.amazon.com/s3/object/teton-cdn?region=us-west-2&prefix=smartslice-cura-plugin/dist/smartslice-cura-r21.1-sso.15.curapackage).

dougkenik:
When I click on "Don't have an account" on the login screen, it goes to the login page, not the registration page. Also, the "Sign in with Ultimaker" button is very blurry

dougkenik (changes requested):
See comment

bradymadams:
> When I click on "Don't have an account" on the login screen, it goes to the login page, not the registration page. Also, the "Sign in with Ultimaker" button is very blurry.

I made the decision to take them to login, because it's actually not very easy to take them straight to registration because of the how the redirects work. I'll look into it a little bit, but I'm not immediately sure how to make it work reliably.

The icons are in the API code, so I'd rather handle any issues with the web forms in the related PR, but I am curious about this because it doesn't look blurry to me. Can you show me a screenshot? Maybe try a hard refresh with CTRL+F5.

image

dougkenik:
> > When I click on "Don't have an account" on the login screen, it goes to the login page, not the registration page. Also, the "Sign in with Ultimaker" button is very blurry.
> 
> I made the decision to take them to login, because it's actually not very easy to take them straight to registration because of the how the redirects work. I'll look into it a little bit, but I'm not immediately sure how to make it work reliably.
> 
> The icons are in the API code, so I'd rather handle any issues with the web forms in the related PR, but I am curious about this because it doesn't look blurry to me. Can you show me a screenshot? Maybe try a hard refresh with CTRL+F5.
> 
> image

image

bradymadams:
I guess our definitions of blurry are different. All I did was take the png you provided me and scaled it down to 200px width. If you want to create new pngs at that size then I'll swap them out.

bradymadams:
@dougkenik okay, "Don't have an account" goes to register. The errant error message is fixed, and the SVGs are in place.

https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/smartslice-cura-r21.1-sso.16.curapackage

dougkenik:
Workflow is clean. Images look great. 2 suggestions:

1. Can we move the warning for remaining trial period to the bottom of the screen?
2. Can we move the "Upgrade" text in the warning underneath the description and have it right aligned?

bradymadams:
Those changes made and also merged the new reminders with the existing subscription status message boxes.

https://teton-cdn.s3-us-west-2.amazonaws.com/smartslice-cura-plugin/dist/smartslice-cura-r21.1-sso.17.curapackage

dougkenik (approved):

AijunHall (approved):
Read through the code to the best of my ability and it looks good to me. The build functionality is good as well.

A few comments about the UI/workflow from testing the build. I'm sure these things are easier said than done so take these with a grain of salt. Also if we are already aware of these, disregard this.
 
When trying to log in with an Ultimaker account, using the same email as a SmartSlice account this page is brought up. Could we create a button here instead of having users enter in the url themselves?
image

When users have gone through the auth and are ready to go into SmartSlice, is this cancel button necessary? I think it would be cleaner and more straightforward if it was only the Ok button.
image

Maybe we could have this page auto delete itself after a couple of seconds for ease of use?
image

bradymadams:
> When trying to log in with an Ultimaker account, using the same email as a SmartSlice account this page is brought up. Could we create a button here instead of having users enter in the url themselves?

I plan on cleaning up some of these pages, but those pages are served from our API (a separate repo), so they're not really the focus of the PR, but your point is noted and I'll look into it when I get back into that code.

> When users have gone through the auth and are ready to go into SmartSlice, is this cancel button necessary? I think it would be cleaner and more straightforward if it was only the Ok button.

I think we should keep it. It provides an explicit action for not granting access to the plug-in. This is a true OAuth workflow, so this could end up being used by 3rd party applications and the user might decide to _not_ continue with authorization. The "Cancel" gives them an obvious action to deny access.

> Maybe we could have this page auto delete itself after a couple of seconds for ease of use?

By "auto delete" do you mean close itself? If so, it's unfortunately [not possible in modern browsers](https://developer.mozilla.org/en-US/docs/Web/API/Window/close). However, this page is going to get a lot more useful content added to it, so we'll want it to stay open anyway. The current page is just a temporary placeholder.

PR #583, opened 2021-08-25
R21.1 stratos welcome

dougkenik:
I broke out the PluginMetadata class so it can be re-used across the plugin, and i moved the home and help URL's into the plugin.json so they are configurable. This is just the beginning, but it will set us up to be more flexible for future partners.

As a first cut, I moved the GIFs in the welcome screen and the Links to the tutorials into pugin.json. I have also put the Stratos videos at the below links. The links to the tutorials are in the original ticket in ZenHub

"product_tour": "https://tetonsim.com/hubfs/In-Product%20Content/welcome-stratos.gif",
"tutorial1": "https://tetonsim.com/hubfs/In-Product%20Content/tutorial1-stratos.gif",
"tutorial2": "https://tetonsim.com/hubfs/In-Product%20Content/tutorial2-stratos.gif"


SmartSlicePlugin/SmartSliceURL.py
```diff
@@ -2,50 +2,57 @@
 from PyQt5.QtCore import pyqtProperty
 
 from .stage.SmartSliceStage import SmartSliceStage
+from .SmartSliceMetadata import PluginMetaData
 
 class URLHandler(QObject):
-    def __init__(self, parent=None):
+    def __init__(self, parent=None, metadeta:PluginMetaData=None):
```
bradymadams, line 8: Spelling on `metadeta`

bradymadams (changes requested):
The code looks good to me. What do you think of dynamically building the links based off a single "product" attribute from the plugin.json? Setting all of these links in the build scripts could be a headache, and I think it would be better if we were consistent on our URLs and could do things like:

```python
productTourGifUrl = "https://tetonsim.com/hubfs/In-Product%20Content/welcome-%s.gif" % product

# or 

productTourGifUrl = "https://tetonsim.com/hubfs/In-Product%20Content/%s/welcome.gif" % product
```

SmartSlicePlugin/SmartSliceURL.py
bradymadams: Also formatting should be

```python
def __init__(self, parent=None, metadeta: PluginMetaData = None):
```

PR #592, opened 2021-08-30
switch printers extruder index check

AijunHall:
Another tiny fix here.


dougkenik (approved):

bradymadams (commented):
Which signal is triggering `_onMachineChanged` in this case? It seems like there is a mismatch between the following two variables

`CuraApplication.getInstance().getExtruderManager().activeExtruderIndex`
`CuraApplication.getInstance().getMachineManager().activeMachine.extruderList`

The reason I ask is because maybe there is a more appropriate signal that emits after Cura has updated `activeExtruderIndex`. If not, then this fix looks good.

AijunHall:
> Which signal is triggering _onMachineChanged in this case?

@bradymadams I believe its `printerConnectedStatusChanged`

https://github.com/tetonsim/is-cura-ui/blob/52a458d584be26deb831d9c77e3ccac053d9ed69/SmartSlicePlugin/SmartSlicePropertyHandler.py#L88

https://github.com/tetonsim/is-cura-ui/blob/52a458d584be26deb831d9c77e3ccac053d9ed69/SmartSlicePlugin/SmartSlicePropertyHandler.py#L155-L157

I don't see another appropriate signal in Cura's MachineManager to use for `printerCheck` since other signals are emitted from a wider pool of globalsettings vs just a machine/printer change. To me it seems like other signals would disrupt the signal-connecting-sequence that is going on with `printerCheck`

bradymadams (approved):

PR #600, opened 2021-10-06
R21.1 blade login

dougkenik:
For blade login


SmartSlicePlugin/SmartSliceAPI.py
```diff
@@ -285,6 +290,40 @@ def callback(status: int, resp: Union[pywim.http.thor.ApiResult, pywim.http.thor
...
+            self.firstName = user_auth.user.first_name
+            self._preferences.setPreference(Preferences.Name, self._first_name)
+            self._token = self._client.get_token()
+            self._createTokenFile()
+            self.authState = AuthState.LoggedIn
+            self.newLogin.emit()
+        else:
+            self.authState = AuthState.NotLoggedIn
+            self._token = None
+            self._createTokenFile()
```
bradymadams, line 325: If the returned status isn't 200 we should log the status code and the error message here. I'm sure we'll find it helpful if someone is having trouble logging in and we need to inspect the log. Calling the `_handleThorErrors` function will handle checking all the various different error codes.

SmartSlicePlugin/SmartSliceAPI.py
```diff
@@ -69,6 +69,11 @@ def __init__(self, connector):
         self._url_handler = connector.extension.url_handler
         self._oauth_handler = None
 
+        #BigRep Blade connection for logging in
+        if self._plugin_metadata.client_name == 'blade':
+            blade_plugin = CuraApplication.getInstance().getPluginRegistry().getPluginObject("BladeLicenseChecker")
+            blade_plugin._smart_slice_account.grantCodeReceived.connect(self.bladeLogin)
```
bradymadams, line 75: Will there ever be a situation where BladeLicenseChecker plugin won't be available? If so, we could avoid a crash might checking for None.

Also, string on line 73 should be in double quotes.

SmartSlicePlugin/SmartSliceAPI.py
```diff
@@ -285,6 +290,40 @@ def callback(status: int, resp: Union[pywim.http.thor.ApiResult, pywim.http.thor
...
+            return
+
+        # Retreice the Code Verifier and the Redirect URL
+        blade_license_checker = CuraApplication.getInstance().getCuraAPI().bladeLicenseChecker
+        authenitcation_service = blade_license_checker._smart_slice_account._authorization_service
+        code_verifier = authenitcation_service.verification_code
+        redirect_uri = authenitcation_service._settings.CALLBACK_URL
+
+        if not code_verifier or not redirect_uri:
+            return
```
bradymadams, line 305: Should we log this condition if it occurs in case this unexpectedly happens and we need to troubleshoot a user's login problems? (same with line 295).

SmartSlicePlugin/SmartSliceAPI.py
```diff
@@ -285,6 +290,40 @@ def callback(status: int, resp: Union[pywim.http.thor.ApiResult, pywim.http.thor
...
         return state, error
 
+    # For logging in to SmartSlice with Blade
+    def bladeLogin(self, grant_code: str = None):
+        if grant_code is None:
+            return
+
+        # Retreice the Code Verifier and the Redirect URL
+        blade_license_checker = CuraApplication.getInstance().getCuraAPI().bladeLicenseChecker
+        authenitcation_service = blade_license_checker._smart_slice_account._authorization_service
```
bradymadams, line 300: Misspelling on this variable.

bradymadams (approved):

PR #607, opened 2021-11-19
Remaining subscription days fix

### LINKED ISSUE #606: Remaining subscription time is incorrect
The remaining days on the subscription for the notification message is being calculated as the subscription end minus the start which always yields the total subscription period. It needs to be the end date minus _today's date_.

https://github.com/tetonsim/is-cura-ui/blob/15bde27b022dac708115db460d7d81f631aef943/SmartSlicePlugin/SmartSliceSubscriptionReminder.py#L36

https://github.com/tetonsim/is-cura-ui/blob/15bde27b022dac708115db460d7d81f631aef943/SmartSlicePlugin/SmartSliceSubscriptionReminder.py#L51

bradymadams:
Resolves #606


SmartSlicePlugin/plugin.json
```diff
@@ -12,7 +12,7 @@
     },
     "smartSliceAccount": "https://account.tetonsim.com",
     "smartSliceAuth": {
-        "basicAuth": false,
+        "basicAuth": true,
```
AijunHall, line 15: Do you mean to revert back to basicAuth? Just asking in case this is leftover from your dev setup.

bradymadams: Yeah, it was intentional because we're going to retain basic auth for the next release. However, after thinking about it. I'm going to revert it. The build system will change this appropriately and I'd rather keep oauth as the default. It makes dev easier, IMO.

AijunHall (approved):

dougkenik (approved):

PR #615, opened 2021-12-14
trademarks

dougkenik:
Updated the logo to include trademarks, and add a TM to the Welcome screen


dougkenik:
Note that we cannot add a trademark to every instance of SmartSlice due to limitations in QML. This should suffice for what we need near term

AijunHall (approved):

bradymadams (approved):

Issue #616, opened 2021-12-14
Load arrows flip direction after SLT orientation is changed and .3mf is saved and reopened


Tyler-Heckenlively:
Anytime an STL is rotated from its original position the load arrows will not maintain the proper direction after the part has been saved as an .3MF. When the .3mf is saved and reopened the load direction will be flipped from the original selected direction.

Link to video demonstration:
https://tetoncompositescom.sharepoint.com/:v:/r/sites/TetonSimulation/Shared%20Documents/Staff/Interns/Tyler/SmartSlice%20Load%20direction%20flip.mp4?csf=1&web=1&e=1uueo0

AijunHall:
Could be relate to #459

Tyler-Heckenlively:
.3MF for model used in example: https://tetoncompositescom.sharepoint.com/:u:/r/sites/TetonSimulation/Shared%20Documents/R%26D/Models/Switch%20Box/UMS5_Switch%20Box.3mf?csf=1&web=1&e=dIfuh4

dougkenik:
This only occurs when the load arrow is saved in a "flipped" state

PR #617, opened 2021-12-17
fix for 616

dougkenik:
This is a fix for #616. We were rotating the origin on project save, which was not being rotated back on project open. We check the origins when we decide to flip the load or not.


bradymadams (approved):

AijunHall (approved):
👍

Issue #605, opened 2021-11-04
Saved validate results are reset upon opening 3mf in Stratos


rickdalgarno:
I used Stratos to run a validation and then saved the results in the Stratos 3mf. If I open a new instance of Stratos and open the saved 3mf, the results from the validation are reset. If, within the same instance, I open the same 3mf again, the validation results are preserved.

Expected: Upon opening a 3mf with saved SmartSlice results, all results are preserved.

Video showing this workflow and 3mf files are here: 

Notes:

- I ran this in Stratos 1.1.0 and 1.2.2 and observed the same behavior.

- When opening a 3mf with saved optimization results, the SmartSlice results are preserved, but the following warning appears. This suggests Stratos or SmartSlice is automatically updating a setting (or something else) which triggers the reset. For a validation, we don't prompt the user to accept the change. For an optimization, we do prompt the user to accept the change.

![image](https://user-images.githubusercontent.com/34459028/140427084-92080f22-e3ee-47ea-b343-c4175ed762b2.png)

Issue #595, opened 2021-09-03
Validate results are not saved in Stratos


rickdalgarno:
Opening a Stratos 3mf that contains results from a validation can produce an error where, upon opening, the validation results are not shown.

Steps to reproduce (see video):

1. Open attached STL.
2. Setup loads and anchors and requirements as shown in video (or in the attached 3mf).
3. Run validate. Save project.
4. Close and reopen Stratos.
5. Open saved file.
6. Note that the Validate results are not shown. Expected outcome is that the saved validate results are displayed.

Notes:

- I can reproduce this in 1.0.0 and 1.1.0 with this specific part and workflow.
- This issue is very sensitive to things like the STL, the number of surfaces selected, and the load type. The workflow shown in this video, however, always produces the error.

[EW50_lever_1p1p0_validate.zip](https://github.com/tetonsim/is-cura-ui/files/7108576/EW50_lever_1p1p0_validate.zip)

https://user-images.githubusercontent.com/34459028/132067228-3b045766-534b-4658-ba68-ad1b5a6ac0d2.mp4

AijunHall:
I can confirm this bug occurs in Cura as well now.

PR #620, opened 2022-03-01
Saved Results Bugfix

AijunHall:
This should take care of #605 and #595

This bug seemed to have been caused specifically by the Factor of Safety/Max Deflection properties

When a user loaded a validated project file, over dimensioned or under dimensioned, the `_onRequirementToolPropertyChanged` signal triggers a change in the CloudStatus back to `ReadyToVerify` upon the first load up action, and therefore hides the results.

I debated on different areas to place a fix (PropertyHandler, CloudStatus, Results updating). I think this is the most appropriate, but I'm open to suggestions.


rickdalgarno:
@AijunHall I'm following the steps in #595 and the issue has not been fixed. I can do a screen share with you when you have time to show you what I am doing.

rickdalgarno:
@bradymadams I manually installed the curapackage you linked in an earlier comment and it shows version 22.0.6 in Stratos now. However, the issue still persists. I save validation results, close and re-open Stratos, then open the saved 3mf but the validation results are not shown.

AijunHall:
Been doing some testing between my system and Ricks. There seems to be alot of inconsistency in how this bug/bugfix occur between different Stratos version. I'll do some more testing and bring some of my findings to you @bradymadams

